PR #6, opened 2017-07-07
Add Bitstamp Ticker/Market and rspec

Swingcloud:


tmlee:
Hi @Swingcloud thank you for your contribution! Before I accept this PR, we need to work on a way to handle "explicit" pairs.

Something along the following line
```
# 1st Precendence
      # Reads a <exchange>.yml file included by the Gem user
      # 2nd Precedence
      # Reads a <exchange>.yml file included within the Gem exchange
      # 3rd Precedence
      # Reads the fetch method (typically an API call)
```

I am facing the same scenario when implementing `korbit` and i think @dominicwong617  had the same thing for `coincheck`, we should aim to get a solution for this quickly before we add more exchanges that falls into this same scenario. I will propose a draft soon for this one

Swingcloud:
sure, if anything needs to change, just tell me!

tmlee:
Sure thing, will keep you posted on that!

tmlee:
@Swingcloud I've pushed the commit in regards to the defined pairs over at https://github.com/coingecko/cryptoexchange/commit/063472a91f3b7ca4398411e837eb0d922a056ae8

The points are:-
```
1. Move the currencies list from lib/cryptoexchange/exchanges/bitstamp/market.rb into a bitstamp.yml file, you can refer to korbit.yml and coincheck.yml on how to define the pairs

2. Under lib/cryptoexchange/exchanges/bitstamp/services/pairs.rb, you want to define a MARKET constant just like how it works for korbit and coincheck. You want to use fetch and inherit from the super class since thats where the fallback is taken care of. Then use adapt method to place the code you have.

3. Refer to korbit/coincheck if in doubt, they are the same scenario
```

This also allows the gem user to create a custom `bitstamp.yml` file to update the currency list so we dont have to always use a PR for new pairs.

Swingcloud:
sure! will do

tmlee:
Hi @Swingcloud thank you for your contributions of several exchanges, totally appreciate them!
Just did a commit on some changes on how the adapt would work, with a Client class, and the usage of an LRUCache. Noticed that some exchanges returns an API with multiple ticker, and the current approach would expose the same interface while not making multiple calls to the same API for each pairs.

If you would rebase from `master`, and make the appropriate changes according to the structure of the exchange, with spec passing. We can start merging all of the pull requests!

Thanks 👍

tmlee:
https://github.com/coingecko/cryptoexchange/pull/12 Might be a useful reference for you

Swingcloud:
yes I will start to change the PR and when PRs are ready, I will let you know!

Swingcloud:
seems like there is some issue on Gatecoin's server, so test doesn't pass, maybe later can restart a test and check whether it will pass or not? can i restart travis ci by myself? @tmlee

lib/cryptoexchange/exchanges/bitstamp/services/market.rb
```diff
@@ -0,0 +1,40 @@
...
+        end
+
+        def ticker_url(market_pair)
+          base = market_pair.base.downcase
+          target = market_pair.target.downcase
+          "#{Cryptoexchange::Exchanges::Bitstamp::Market::API_URL}/#{base}#{target}"
+        end
+
+        def adapt(output, market_pair)
+          ticker = Coincheck::Models::Ticker.new
```
dominicwong617: @Swingcloud `Bitstamp` not `Coincheck`

lib/cryptoexchange/exchanges/bitstamp/services/market.rb
```diff
@@ -0,0 +1,40 @@
...
+          ticker.base = market_pair.base
+          ticker.target = market_pair.target
+          ticker.market = Bitstamp::Market::NAME
+          ticker.ask = output['ask']
+          ticker.bid = output['bid']
+          ticker.last = output['last']
+          ticker.high = output['high']
+          ticker.low = output['low']
+          ticker.volume = output['volume'].to_f
```
dominicwong617: @tmlee @Swingcloud should we use `float` or `BigDecimal`?

lib/cryptoexchange/exchanges/bitstamp/services/pairs.rb
```diff
@@ -0,0 +1,26 @@
+module Cryptoexchange::Exchanges
+  module Bitstamp
+    module Services
+      class Pairs < Cryptoexchange::Services::Pairs
+        MARKET = Bitstamp::Market
```
dominicwong617: @Swingcloud is this needed?

lib/cryptoexchange/exchanges/bitstamp/services/pairs.rb
```diff
@@ -0,0 +1,26 @@
...
+
+        def fetch
+          output = super
+          adapt(output)
+        end
+
+        def adapt(output)
+          market_pairs = []
+          output.each do |pair|
+            market_pairs << Bitstamp::Models::MarketPair.new(
```
dominicwong617: @Swingcloud appreciate if you can nest the args righ around `Bitstamp::Models::MarketPair`, thanks

spec/exchanges/bitstamp/market_spec.rb
```diff
@@ -0,0 +1,6 @@
+require 'spec_helper'
+
+RSpec.describe Cryptoexchange::Exchanges::Bitstamp::Market do
+  it { expect(Cryptoexchange::Exchanges::Bitstamp::Market::NAME).to eq 'bitstamp' }
```
dominicwong617: @Swingcloud I guess you can use `described_class` here and everywhere

lib/cryptoexchange/exchanges/bitstamp/services/market.rb
Swingcloud: ok

lib/cryptoexchange/exchanges/bitstamp/services/pairs.rb
Swingcloud: maybe because in `Cryptoexchange::Services::Pairs::MARKET` is nil?

lib/cryptoexchange/exchanges/bitstamp/services/pairs.rb
Swingcloud: will do

spec/exchanges/bitstamp/market_spec.rb
Swingcloud: ok

lib/cryptoexchange/exchanges/bitstamp/services/market.rb
tmlee: I tend to use them interchangeable. What's the underlying difference?
Would think `BigDecimal` does no harm

lib/cryptoexchange/exchanges/bitstamp/services/pairs.rb
tmlee: @dominicwong617 open to suggestion here.

This is used to determine whether to grab pairs via a URL, or fallback to YML

If the `market_url` is defined, it will take the URL. Otherwise it will try to read a YML.
That `MARKET` name is used to find the `yml` filename. If we have a better way to propagate the exchange name to the superclass, that would be better

tmlee: One possible improvement is to use the classname string and use that to find the yml

dominicwong617: Actually I was just thinking why we need to define a constant as another constant, that's about it. 

And actually I think we don't need to hard code markets since we can grab the class name (as you say) and transform it (`underscore` it or something like that)

tmlee: https://github.com/coingecko/cryptoexchange/pull/13

dominicwong617: but it encourages a standard

dominicwong617:
@Swingcloud you have a conflict

dominicwong617:
@tmlee merging this

PR #7, opened 2017-07-07
Add Livecoin + LRUTTLCache to cache the response

tmlee:


lib/cryptoexchange/services/market.rb
```diff
@@ -2,9 +2,15 @@ module Cryptoexchange
   module Services
     class Market
       def fetch(endpoint)
-        fetch_response = HTTP.get(endpoint)
+        fetch_response = http_get(endpoint)
         response = JSON.parse(fetch_response.to_s)
       end
+
+      def http_get(endpoint)
```
dominicwong617: @tmlee why do you need a method to do `HTTP.get(endpoint)`?

PR #8, opened 2017-07-07
add Anx

Swingcloud:


tmlee:
If the API supports market pair (ie. being able to grab a list of pairs via an API), no need to implement the `yml` file. `yml` file would be for exchange that does not support such API, and we have to supply a constant; with the option for Gem user to override it too.

Swingcloud:
sure! will do

tmlee:
@Swingcloud I am sorry, i thought earlier i saw a url in the `pairs` class; i think this is correct!

lib/cryptoexchange/exchanges/anx/services/pairs.rb
```diff
@@ -0,0 +1,26 @@
...
+          adapt(output)
+        end
+
+        def adapt(output)
+          market_pairs = []
+          output.each do |pair|
+            market_pairs << Anx::Models::MarketPair.new(
+                              base: pair[:base],
+                              target: pair[:target],
+                              market: MARKET::NAME
```
dominicwong617: @Swingcloud change this to `Anx::Market::NAME`

dominicwong617: everything else looks good

spec/exchanges/anx/services/pairs_spec.rb
```diff
@@ -0,0 +1,4 @@
+require 'spec_helper'
+
+RSpec.describe Cryptoexchange::Exchanges::Anx::Services::Pairs do
```
dominicwong617: @Swingcloud why is this empty?

dominicwong617:
@tmlee merging this

PR #12, opened 2017-07-11
Add BitcoinIndonesia

tmlee:


lib/cryptoexchange/exchanges/bitcoin_indonesia/services/market.rb
```diff
@@ -0,0 +1,41 @@
...
+          ticker = BitcoinIndonesia::Models::Ticker.new
+          ticker.base = market_pair.base
+          ticker.target = market_pair.target
+          ticker.market = BitcoinIndonesia::Market::NAME
+          ticker.bid = output['buy']
+          ticker.ask = output['sell']
+          ticker.last = output['last']
+          ticker.high = output['high']
+          ticker.low = output['low']
+          ticker.volume = output["vol_#{ticker.base.downcase}"].to_f
```
tmlee: Actually maybe BigDecimal would be safer and sensible

lib/cryptoexchange/exchanges/bitcoin_indonesia/bitcoin_indonesia.yml
```diff
@@ -0,0 +1,19 @@
+:pairs:
```
dominicwong617: @tmlee not that it matters, but do we need the extra colon in front? less keystrokes bro....

lib/cryptoexchange/exchanges/bitcoin_indonesia/services/market.rb
dominicwong617: @tmlee yes let's use BigDecimal for all exclusively

tmlee: Noted lets create a separate pull request for that

lib/cryptoexchange/exchanges/bitcoin_indonesia/bitcoin_indonesia.yml
tmlee: I want to access the keys using `:`, any better ideas? Write custom methods? The helper methods are only available in Rails package i think

PR #17, opened 2017-07-13
Implement Novaexchange

yihangho:


lib/cryptoexchange/exchanges/novaexchange/models/market_pair.rb
```diff
@@ -0,0 +1,14 @@
+module Cryptoexchange::Exchanges
+  module Novaexchange
+    module Models
+      class MarketPair < Cryptoexchange::Models::MarketPair
+        attr_accessor :name
+
+        def initialize(params={})
+          super
+          @name = params[:name]
```
dominicwong617: @yihangho what is this for?

yihangho: This is used to query ticker for individual trading pair (see [here][ticker-url]). For most (but not all) trading pairs, the name is just `#{base}_#{target}`. So I'm using this to keep track of its name.

[ticker-url]: https://github.com/yihangho/cryptoexchange/blob/777720db00a4a584d1e933aff3044cd63740483b/lib/cryptoexchange/exchanges/novaexchange/services/market.rb#L19

dominicwong617: is it possible to achieve the same result without introducing new instance variables?

yihangho: Well, the easy way out is to hardcode this special case in the code, but that's very hacky.

dominicwong617: Which trading pair does not have the `base_target` format?

yihangho: ESP/DOGE - its name is ESP_BLRY. @tmlee thinks it's likely a bug on their side, which I agree. However, querying ESP_DOGE gives a 404.

tmlee: @dominicwong617 seems like a bug to me, the pair name is `ESP_BLRY`, and within the hash, the base is `DOGE`

dominicwong617: I think it's better to do it as a `hack` with a `FIXME` and `Issue` on GitHub so that we remember to take it out later

tmlee: @dominicwong617 will you be able to do that and merge this PR?

dominicwong617: ok

dominicwong617:
@yihangho looks good

PR #19, opened 2017-07-13
BigDecimal for prices, Integer (Numeric) for timestamp

dominicwong617:


tmlee:
I know using the conditional `?` makes it safe.
But are we being too paranoid here and is there a better way to make it work? 
Worst case we should refactor this out into a Condition Handling class.
Though i really wished we can just get away with `ticker['data'].to_d` if possible.

But if we really have to do these checks, I am not too agreeable with having it all sprinkled around the codebase. Then should we ask ourself, what is with retain using `Float` instead of `BigDecimal`

dominicwong617:
@tmlee 

1. I am more than happy to refactor it into a condition handling thing (probably just a helper method). Just wasn't aware of the repetition when I did the mechanical work.
2. The program must work if we want people to use it, so safety is pretty basic.
3. Accuracy is important as we don't know what people do with the numbers, especially when they are prices. I can't agree on using float.

dominicwong617:
@tmlee I added a helper method

lib/cryptoexchange/exchanges/bitcoin_indonesia/services/market.rb
```diff
@@ -20,19 +20,19 @@ def ticker_url(market_pair)
...
+          ticker           = BitcoinIndonesia::Models::Ticker.new
+          ticker.base      = market_pair.base
+          ticker.target    = market_pair.target
+          ticker.market    = BitcoinIndonesia::Market::NAME
+          ticker.bid       = NumericHelper.to_d(output['buy'])
+          ticker.ask       = NumericHelper.to_d(output['sell'])
+          ticker.last      = NumericHelper.to_d(output['last'])
+          ticker.high      = NumericHelper.to_d(output['high'])
+          ticker.low       = NumericHelper.to_d(output['low'])
+          ticker.volume    = output["vol_#{ticker.base.downcase}"] ? BigDecimal.new(output["vol_#{ticker.base.downcase}"].to_s) : nil
```
tmlee: This needs to use the Helper

tmlee:
Looks good to me except for the one missing Helper usage

tmlee:
Fix that and this can be merged

dominicwong617:
@tmlee ok I will fix it later and start merging other PRs

Swingcloud:
@tmlee this is good to merge!

PR #18, opened 2017-07-13
Lakebtc

dominicwong617:

PR #20, opened 2017-07-14
Fix test for Nova Exchange

yihangho:
Should fix the breakage of #18.


tmlee:
Thanks, I think for integration test we can do a `nil` check.
For actual value assignment, a unit test with stubbed results makes more sense.

yihangho:
Yeah, I forgot to swap it out with a `nil` check when I was referring to the tests from other exchanges.

PR #22, opened 2017-07-15
Implement c-cex.com

yihangho:


lib/cryptoexchange/exchanges/ccex/services/market.rb
```diff
@@ -0,0 +1,50 @@
...
+          volumes_output = super(volumes_url(market_pair))
+          adapt(ticker_output, volumes_output, market_pair)
+        end
+
+        def ticker_url(market_pair)
+          name = "#{market_pair.base}-#{market_pair.target}".downcase
+          "#{Cryptoexchange::Exchanges::Ccex::Market::API_URL}/#{name}.json"
+        end
+
+        def volumes_url(market_pair)
```
dominicwong617: @yihangho sorry to be picky, but let's make it `volume_url`

dominicwong617: and `volume_output`

lib/cryptoexchange/exchanges/ccex/services/market.rb
```diff
@@ -0,0 +1,50 @@
...
+
+        def adapt(ticker_output, volumes_output, market_pair)
+          market = ticker_output['ticker']
+          volume = HashHelper.dig(volumes_output, 'ticker', market_pair.base.downcase, 'vol')
+
+          ticker = Ccex::Models::Ticker.new
+          ticker.base      = market_pair.base
+          ticker.target    = market_pair.target
+          ticker.market    = Ccex::Market::NAME
+          ticker.last      = market['lastprice'] ? BigDecimal.new(market['lastprice'].to_s) : nil
```
dominicwong617: @yihangho I am thinking maybe we can have a helper method that does this

lib/cryptoexchange/helpers/hash_helper.rb
```diff
@@ -0,0 +1,20 @@
...
+    # "Polyfill" of the Hash#dig method from Ruby 2.3
+    #
+    #   Examples
+    #
+    #     > dig({a: {b: 1}}, :a, :b)
+    #     => 2
+    #
+    #     > dig({a: {b: 1}}, :a, :c)
+    #     => nil
+    def dig(obj, *args)
```
dominicwong617: @yihangho can we upgrade to Ruby 2.4?

@tmlee what do you think?

lib/cryptoexchange/exchanges/ccex/services/market.rb
yihangho: Will do.

lib/cryptoexchange/exchanges/ccex/services/market.rb
yihangho: You mean the conversion from string/float/int/nil to big decimal? Yup, I think it's a good idea.

yihangho:
Test is failing, but it should be fixed when #22 is merged. I'll rebase again when it's merged.

tmlee:
@yihangho https://github.com/coingecko/cryptoexchange/pull/22 is this PR yes? Or are you referring to a different PR

yihangho:
@tmlee Oh, I made a typo, should be #23

lib/cryptoexchange/exchanges/ccex/services/market.rb
dominicwong617: @yihangho can you use `NumericHelper.to_d`?

dominicwong617: it's on `master`

yihangho:
The tests are failing because of connectivity issues with c-cex. Will try again in a few hours.

lib/cryptoexchange/helpers/hash_helper.rb
dominicwong617: @yihangho @tmlee Ping

yihangho: @dominicwong617 @tmlee I'm on the fence for this one. On one hand, I think we should keep Ruby 2.2 because as a library, we should be as flexible as we reasonably can be, especially in the context of external dependencies (Ruby included). In this case, I think simple polyfills like this are reasonable. On the other hand, Ruby 2.2 is actually quite old now and Ruby core [is not supporting it][1] anymore.

[1]: https://www.ruby-lang.org/en/news/2017/03/28/ruby-2-2-7-released/

dominicwong617: @yihangho Fair enough, although I would argue that this is really just a helper method instead of a polyfill (not important though).

dominicwong617:
@yihangho there are conflicts.

lib/cryptoexchange/helpers/hash_helper.rb
yihangho: @dominicwong617 Yeah, I agree. I decided against making it a polyfill because I prefer not touching the global scope, especially my implementation isn't 100% compatible with Hash#dig.

tmlee: What is the lowest version of Ruby where `dig` is supported? `2.3` yes?
I would prefer to keep `2.2` for now if possible, same argument which is to support as wide as possible until it is a problem that we need to bump to 2.3
However, forcing 2.3 earlier would also mean we do not need to backward compatible a lot of things later on when they accumulate.

dominicwong617: @tmlee actually you're correct, my blurry memory told me it was `2.4` but it is actually `2.3`. Going to merge this one

PR #23, opened 2017-07-17
implement Gemini (ready)

Swingcloud:
seems like novaexchange have fixed the ESP_DOGE issue, and there is no more ESP in market pairs

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ Or install it yourself as:
 | Bitstamp          | Y       |            |         |         | User-Defined|
 | Bittrex           |         |            |         |         |             |
 | GDAX              |         |            |         |         |             |
-| Gemini            |         |            |         |         |             |
+| Gemini            | Y       |            |         |         | Y           |
 | Kraken            |         |            |         |         |             |
 | Poloniex          |         |            |         |         |             |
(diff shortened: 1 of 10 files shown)


spec/exchanges/novaexchange/integration/market_spec.rb
```diff
@@ -29,21 +29,4 @@
     it { expect(@ticker.timestamp).to_not be nil }
     it { expect(@ticker.payload).to_not be nil }
   end
-
```
dominicwong617: @Swingcloud 

1. they should be in 2 different PRs (but it's fine this time)
2. did you remove the hack in the actual code?

yihangho: @dominicwong617 I did a quick check of the current response from Nova, indeed it has been removed, but they added another edge case: https://novaexchange.com/remote/v2/market/info/ESP2_DONATE_/. It seems like it's not an active market, should be safe to ignore.

Swingcloud: yeah I've remove the hack in actual code

tmlee: i noticed that too and issued a commit for that. Just noticed the discussion here later on

PR #24, opened 2017-07-17
Remove unused params (ready)

Swingcloud:
when merging #23, rspec error will be solved


dominicwong617:
@Swingcloud tests are failing

Swingcloud:
yeah it's because special case on novaexchange no longer exist, when PR #23 merge, the test will pass

PR #28, opened 2017-07-18
Implement Bittrex

Swingcloud:


lib/cryptoexchange/exchanges/bittrex/services/market.rb
```diff
@@ -0,0 +1,42 @@
...
+        end
+
+        def adapt(output)
+          ticker = Bittrex::Models::Ticker.new
+          result = output['result'][0]
+
+          ticker.base      = result['MarketName'].split("-")[0]
+          ticker.target    = result['MarketName'].split("-")[1]
+          ticker.market    = Bittrex::Market::NAME
+          ticker.last      = result['Last'] ? BigDecimal.new(result['Last'].to_s) : nil
```
dominicwong617: @Swingcloud  can we use `NumericHelper.to_d` from `master`?

lib/cryptoexchange/exchanges/bittrex/services/market.rb
```diff
@@ -0,0 +1,42 @@
...
+
+          ticker.base      = market['MarketName'].split("-")[0]
+          ticker.target    = market['MarketName'].split("-")[1]
+          ticker.market    = Bittrex::Market::NAME
+          ticker.last      = NumericHelper.to_d(market['Last'])
+          ticker.high      = NumericHelper.to_d(market['High'])
+          ticker.low       = NumericHelper.to_d(market['Low'])
+          ticker.ask       = NumericHelper.to_d(market['Ask'])
+          ticker.bid       = NumericHelper.to_d(market['Bid'])
+          ticker.volume    = NumericHelper.to_d(market['BaseVolume']) ? BigDecimal.new(market['BaseVolume'].to_s) : nil
```
dominicwong617: @Swingcloud that's a careless mistake 😂

spec/client_spec.rb
```diff
@@ -20,6 +20,7 @@
                             liqui
                             gemini
                             hitbtc
+                            bittrex
```
dominicwong617: @Swingcloud actually the real way to test this is to arrange the order yourself manually without calling `sort` because you're calling `sort` on the actual method as well. If `sort` isn't behaving the way you want, you wouldn't have known because the test will still pass.

lib/cryptoexchange/exchanges/bittrex/services/market.rb
Swingcloud: 😂 sorry

spec/client_spec.rb
Swingcloud: yeah I think I'll take the @tmlee mentioned in Gitter, just stub it and see that method works or not

PR #31, opened 2017-07-18
Fix base/target pair for Nova

yihangho:


tmlee:
Thanks @yihangho change looks good, but will you be able:-

1.  Include an elaboration describing the fix in the commit message
2. Some comments in there describing why it is intuitive

This is to ensure of Novaexchange ever make changes to the API, we know why and where to look for in the future.

yihangho:
@tmlee Will do. 

I think we should add some simple sanity test in the test suite too, since it's obvious that this is a bug on their side. For e.g., I propose in the integration test, we fetch the ticker for LTC/BTC and ensure that the price should be < 1. What do you think?

tmlee:
@yihangho Can do, just describe the test in a way that makes sense.
If there are other pairs we can use that would still hold true for `< 1` instead of LTC, that would be a better choice actually. Like some a smaller crypto instead of LTC

yihangho:
> @yihangho Can do, just describe the test in a way that makes sense.
> If there are other pairs we can use that would still hold true for < 1 instead of LTC, that would be a better choice actually. Like some a smaller crypto instead of LTC

@tmlee How about Doge? The current rate is 0.000,000,77.

tmlee:
@yihangho That seems good, lets use that

yihangho:
Cool, it's already updated to use Doge.

tmlee:
👍

PR #33, opened 2017-07-19
change available_exchanges array to method can autoload each exchanges

Swingcloud:
maybe it's a dumb method now, but right now we can just add a folder name than Client can autoload exchange

### DIFF
--- a/lib/cryptoexchange/client.rb
+++ b/lib/cryptoexchange/client.rb
@@ -1,24 +1,10 @@
 module Cryptoexchange
   class Client
-    AVAILABLE_EXCHANGES = %w(
-                            btcc
-                            anx
-                            bitcoin_indonesia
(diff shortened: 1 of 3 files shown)


lib/cryptoexchange/client.rb
```diff
@@ -1,24 +1,10 @@
...
-                            bitstamp
-                            okcoin
-                            novaexchange
-                            liqui
-                            gemini
-                            hitbtc
-                          )
+    class << self
+      def available_exchanges
+        Dir.entries("./lib/cryptoexchange/exchanges")[2..-1].freeze
```
dominicwong617: @Swingcloud this is fine, but can you write a test?

Swingcloud: will do

PR #37, opened 2017-07-20
implement Quoine(ready)

Swingcloud:
Something need to ask, according to Quoine's api doc, there is another api `product/:id` [link](https://developers.quoine.com/?ruby#get-a-product) which has single currency's market data.

But it uses `:id` to show it, since we don't have another columns can save on models `market_pair`, so we just list all market ticker?

### DIFF
--- a/lib/cryptoexchange/exchanges/quoine/market.rb
+++ b/lib/cryptoexchange/exchanges/quoine/market.rb
@@ -1,0 +1,8 @@
+module Cryptoexchange::Exchanges
+  module Quoine
+    class Market
+      NAME = 'quoine'
+      API_URL = 'https://api.quoine.com'
+    end
+  end
+end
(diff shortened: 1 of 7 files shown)


tmlee:
@Swingcloud My view is that, if the API supports `single_call` for multiple tickers, we should always try to use that.

Unless the API has a rate limit on the multiple tickers single call, then we would consider using the single ticker. This is because the single call is most optimal.

In a case where we need to use a single API ticker call and in a situation like this where `:id` needs to be passed in. We can add the extra column in the subclasses `market_pair` class. (Although we had a separate discussion somewhere about doing away with this since we have not reached a need to subclass and customize yet)

Swingcloud:
@tmlee  oh sorry, this is the previous question about this PR, I already call the multiple tickers on this exchanges. but still appreciate your answer!!! very helpful although I don't have sufficient knowledge about this.

tmlee:
No worries. We will know when we hit a rate limit error. Better to use the single API call whenever we can.

PR #39, opened 2017-07-23
change client rspec, don't need to add services once added

Swingcloud:
use stub to make sure this could list a dir name

### DIFF
--- a/spec/client_spec.rb
+++ b/spec/client_spec.rb
@@ -1,30 +1,17 @@
 require 'spec_helper'
 
 RSpec.describe Cryptoexchange::Client do
-  it 'loads every available exchanges according to folder name' do
-    available_exchanges = %w(
-                            btcc
-                            anx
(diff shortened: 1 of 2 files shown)


spec/client_spec.rb
```diff
@@ -2,29 +2,7 @@
...
-                            liqui
-                            gemini
-                            hitbtc
-                            bithumb
-                            ether_delta
-                          ).sort
-    expect(described_class.available_exchanges).to eq available_exchanges
-
+    allow(described_class).to receive(:available_exchanges).and_return('["anx","bitcoin_indonesia"]')
+    expect(described_class.available_exchanges).to eq('["anx","bitcoin_indonesia"]')
```
dominicwong617: @Swingcloud this does not test anything

Swingcloud:
if I check all available exchanges can load pair, but if you type wrong service name will show name error, is that okay? @dominicwong617

PR #42, opened 2017-07-27
implement Gemini order book

Swingcloud:
right now just list all hashes in `bids` and `asks`, is there any better way to list that?

### DIFF
--- a/lib/cryptoexchange/client.rb
+++ b/lib/cryptoexchange/client.rb
@@ -25,5 +25,21 @@ module Cryptoexchange
         end
       end
     end
+
+    def order_book(market_pair)
+      exchange = market_pair.market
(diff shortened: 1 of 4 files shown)


tmlee:
Good start. We need to standardize the output into a Class. Let me see what we can do about this

tmlee:
Can probably cherry-pick this commit https://github.com/tmlee/cryptoexchange/commit/597884cebeb29f02c915c49e7ab7970e198a96d1

tmlee:
@Swingcloud I will try to merge this in

tmlee:
Added https://github.com/coingecko/cryptoexchange/commit/cb92bd844500af16febd66c7336a57a8f013cfe4 on top of this

PR #43, opened 2017-07-27
Implement Yobit Ticker

tmlee:


lib/cryptoexchange/exchanges/yobit/models/market_pair.rb
```diff
@@ -0,0 +1,8 @@
+module Cryptoexchange::Exchanges
+  module Yobit
+    module Models
+      class MarketPair < Cryptoexchange::Models::MarketPair
```
dominicwong617: @tmlee please do not create child model classes from now on, thanks.

PR #52, opened 2017-07-28
implement bitfinex(ready)

Swingcloud:


lib/cryptoexchange/exchanges/bitfinex/models/market_pair.rb
```diff
@@ -0,0 +1,8 @@
+module Cryptoexchange::Exchanges
+  module Bitfinex
+    module Models
+      class MarketPair < Cryptoexchange::Models::MarketPair
```
dominicwong617: @Swingcloud please do not create child model classes from now on, thanks.

Swingcloud: will do

dominicwong617:
I am thinking for other tests we also should do `is a Numeric` instead of `not nil`
@Swingcloud

Issue #58, opened 2017-08-01
Novaexchange volume glitch


tmlee:
@yihangho We implemented novaexchange to have the base/target switched.
However the novaexchange API returns the volume based on their API base

For instance

![screen shot 2017-08-02 at 1 39 50 am](https://user-images.githubusercontent.com/110264/28838667-96d6821e-7723-11e7-98c9-00f4692273d8.png)

and the API results
![screen shot 2017-08-02 at 1 39 45 am](https://user-images.githubusercontent.com/110264/28838671-97f33f70-7723-11e7-858e-079874a4af99.png)

On their API pair of BTC_ETH which translates to base of ETH and target BTC.
The volume needs to be in terms of ETH, however the API returns in BTC terms

tmlee:
Fix with https://github.com/coingecko/cryptoexchange/pull/73/files

tmlee:
PR merged

PR #68, opened 2017-08-03
change ticker volume to base volume(ready)

Swingcloud:


tmlee:
For `coin_exchange`, the 2 volume attributes seem to return the same data
https://www.coinexchange.io/api/v1/getmarketsummaries which looks like a bug on their API

Since we are looking for base, shouldn't we opt for `Volume` instead of `BTCVolume`?

Swingcloud:
Ok! so i just change `coin_exchange` volume to `Volume` instead of `BTCVolume`?

tmlee:
@Swingcloud That won't be exactly correct too, since the `volume` is denominated in `BTC`.
I dont know if it is correct to divide the `BTC` with the `base` currency.
@dominicwong617 @yihangho thoughts?

tmlee:
Same issue with https://github.com/coingecko/cryptoexchange/issues/58
However I do think the best solution is to divide its own price.

I've a few thoughts that I am not sure which is the best yet.

1. Dictate and divide the `target` volume with the price to get volume in `base` 
2. If `base` volume is not available, then do not assign to the volume. Create a new column like `volume_target`. Give the user the option to use that and decide if they want to divide it.

Swingcloud:
I think divide is the better way to solve this, let me do it

Swingcloud:
maybe option 2 is the better way to deal with this?

tmlee:
@Swingcloud I think lets hotfix this and take care of the volume with the user by calculating it against the `last` price. I'll create a separate issue listing the exchange that contains this issue to revisit.

PR #90, opened 2017-08-20
Small revision

Swingcloud:
some upcase for base price and payload lack of volume

### DIFF
--- a/lib/cryptoexchange/exchanges/ccex/services/market.rb
+++ b/lib/cryptoexchange/exchanges/ccex/services/market.rb
@@ -39,7 +39,7 @@ module Cryptoexchange::Exchanges
           ticker.low       = NumericHelper.to_d(market['low'])
           ticker.volume    = NumericHelper.to_d(volume)
           ticker.timestamp = DateTime.now.to_time.to_i
-          ticker.payload   = market
+          ticker.payload   = market.merge({ "volume" => volume })
           ticker
         end
       end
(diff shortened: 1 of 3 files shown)


lib/cryptoexchange/exchanges/gemini/services/pairs.rb
```diff
@@ -15,8 +15,8 @@ def adapt(output)
             base =
             target = pair[3..-1]
             market_pairs << Cryptoexchange::Models::MarketPair.new(
-                              base: pair[0..2],
-                              target: pair[3..-1],
+                              base: pair[0..2].upcase,
```
dominicwong617: @Swingcloud why do you need to do this?

```
module Cryptoexchange
  module Models
    class MarketPair
      attr_accessor :base, :target, :market

      def initialize(params={})
        @base = params[:base]
        @target = params[:target]
        @market = params[:market]
      end

      def target
        @target.to_s.upcase
      end

      def base
        @base.to_s.upcase
      end
    end
  end
end
```

lib/cryptoexchange/exchanges/jubi/services/pairs.rb
```diff
@@ -13,7 +13,7 @@ def adapt(output)
           default_target = 'CNY'
           output.keys.map do |base|
             Cryptoexchange::Models::MarketPair.new({
-              base: base,
+              base: base.upcase,
```
dominicwong617: @Swingcloud and this?

lib/cryptoexchange/exchanges/ccex/services/market.rb
```diff
@@ -39,7 +39,7 @@ def adapt(ticker_output, volumes_output, market_pair)
           ticker.low       = NumericHelper.to_d(market['low'])
           ticker.volume    = NumericHelper.to_d(volume)
           ticker.timestamp = DateTime.now.to_time.to_i
-          ticker.payload   = market
+          ticker.payload   = market.merge({ "volume" => volume })
```
dominicwong617: @Swingcloud I would rather not modify the original payload

PR #91, opened 2017-08-20
implement Bitbay

Swingcloud:
- implement Bitbay pairs and market ticker
- it related to #84 
- [x] I have added Specs


tmlee:
👍 except for the pairs yml file and we are good to go

Swingcloud:
actually Bitbay doesn't have the list of the pairs, so I just use the currencies they show on the index

tmlee:
@Swingcloud Yea, listing is the way to deal with it at the moment if there are no API listing the pairs

PR #92, opened 2017-08-20
implement Bitso

Swingcloud:
- implement Bitso pairs and market ticker
- it related to #83
- [x] I have added Specs


tmlee:
Looks 👍  except for the file name `pairs.rb`

Swingcloud:
😅  my bad

Issue #95, opened 2017-08-22
ticker pass the wrong timestamp 


Swingcloud:
seems like some stickers pass timestamp in millisecond:
- bithumb
- btcc
- chbtc
- gemini
- viabtc

tmlee:
Closed to mark as resolved after merged

PR #96, opened 2017-08-22
[#hotfix] Add timestamp case and revise wrong timestamp

Swingcloud:
- Hotfix some exchange pass timestamp in milliseconds 
- #95 
- [x] I have added Specs


dominicwong617:
@tmlee I am merging this.

Issue #82, opened 2017-08-17
Implement Huobi Ticker


tmlee:

yihangho:
I'm working on this

dominicwong617:
@yihangho heard you have issues with the Chinese doc, I can work on it instead

PR #102, opened 2017-08-23
Implement Huobi

yihangho:
Implement Huobi
#82 
- [x] I have added Specs

### DIFF
--- a/spec/exchanges/huobi/integration/market_spec.rb
+++ b/spec/exchanges/huobi/integration/market_spec.rb
@@ -1,0 +1,32 @@
+require 'spec_helper'
+
+RSpec.describe 'Huobi integration specs' do
+  let(:client) { Cryptoexchange::Client.new }
+
+  it 'fetch pairs' do
+    pairs = client.pairs('huobi')
+    expect(pairs).not_to be_empty
+
+    pair = pairs.first
+    expect(pair.base).not_to be_nil
+    expect(pair.target).not_to be_nil
+    expect(pair.market).to eq 'huobi'
+  end
+
+  it 'fetch ticker' do
+    eth_cny_pair = Cryptoexchange::Models::MarketPair.new(base: 'ETH', target: 'CNY', market: 'huobi')
+    ticker = client.ticker(eth_cny_pair)
+
+    expect(ticker.base).to eq 'ETH'
+    expect(ticker.target).to eq 'CNY'
+    expect(ticker.market).to eq 'huobi'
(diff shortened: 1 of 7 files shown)


spec/exchanges/huobi/integration/market_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+    expect(ticker.base).to eq 'ETH'
+    expect(ticker.target).to eq 'CNY'
+    expect(ticker.market).to eq 'huobi'
+    expect(ticker.last).to be_a Numeric
+    expect(ticker.bid).to be_a Numeric
+    expect(ticker.ask).to be_a Numeric
+    expect(ticker.high).to be_a Numeric
+    expect(ticker.volume).to be_a Numeric
+    expect(ticker.timestamp).to be_a Numeric
+    expect(DateTime.strptime(ticker.timestamp.to_s, '%s').year).to be <= 5000
```
dominicwong617: @yihangho @Swingcloud 
Regardless of whether we should use our own timestamp, let's use `Time.zone.now.year` to test this instead of `5000`

dominicwong617: if you guys agree then we will have to do a global replace.

tmlee: That's fine by me

yihangho: Should I update this PR or will it be part of the global replace?

dominicwong617: @yihangho just change yours

dominicwong617:
@yihangho Actually, what about `https://api.huobi.pro/market` ? There seems to be more than one Huobi API that queries different trading pairs, with BTC as the target currency?

yihangho:
@dominicwong617 Getting 404 for that

dominicwong617:
@yihangho Okay. I am fine with merging the PR then, but could you add some comments (TODOs) so that we might revisit this matter later?

dominicwong617:
@yihangho I will merge this after the comments and `5000` are modified.

yihangho:
@dominicwong617 Actually, I was able to get all the endpoints that I'm using to work with the .pro version. Should we support both or split the .pro one into a separate exchange?

It seems like .com is an exchange that deals with fiat while .pro deals only with crypto.

yihangho:
@dominicwong617 I've updated my PR

dominicwong617:
@yihangho looks good, thanks!

PR #110, opened 2017-08-26
BX Thailand Ticker

dominicwong617:
- What is the purpose of this Pull Request? https://github.com/coingecko/cryptoexchange/issues/106
- [x] I have added Specs


tmlee:
👍  looks good!

PR #115, opened 2017-09-15
Fix Github repo link in Contributing section

alexanderdavidpan:
This fixes the broken link in documentation: http://www.rubydoc.info/gems/cryptoexchange/0.10.2


tmlee:
Good catch. Thanks for the contribution!

Issue #112, opened 2017-09-06
A method to see which exchange supports a specific cryptocurrency


dominicwong617:
For example,

```
client = Cryptoexchange::Client.new
client.exchanges_for('BTC')
['gatecoin', 'anx', ...]
```

Swingcloud:
let me do this

Swingcloud:
NOTE: if endpoint is unavailable, move to another exchange first

PR #120, opened 2017-09-22
Exchange for (Currency) from services

Swingcloud:
- What is the purpose of this Pull Request? add rescue method for possible error for calling api endpoint, and add `exchange_for` and `available_exchange`
- What is the related issue for this Pull Request? #112 
- [ ] I have added Specs
- [ ] (If implementing Market Ticker) I have verified that the `volume` attributes is correct
- [ ] (If implementing Market Ticker) I have verified that the `base` and `target` is correct

this is really slow, cuz it needs to run through all services, takes about 2 mins to do it

### DIFF
--- a/lib/cryptoexchange/client.rb
+++ b/lib/cryptoexchange/client.rb
@@ -36,5 +36,20 @@ module Cryptoexchange
         folder_name.split('/').last
       end
     end
+
+    def exchange_for(currency)
+      exchanges = []
+      available_exchanges.each do |exchange|
+        pairs = pairs(exchange)
+        next if pairs.is_a?(Hash) && !pairs[:error].empty?
+        pairs.each do |pair|
+          if [pair.base, pair.target].include?(currency.upcase)
+            exchanges << exchange
+            break
+          end
+        end
+      end
+      exchanges
+    end
   end
 end
--- a/lib/cryptoexchange/client.rb
+++ b/lib/cryptoexchange/client.rb
@@ -25,5 +25,12 @@ module Cryptoexchange
         end
       end
     end
+
+    def available_exchanges
+      folder_names = Dir[File.join(File.dirname(__dir__), 'cryptoexchange', 'exchanges', '**')]
+      folder_names.map do |folder_name|
+        folder_name.split('/').last
+      end
+    end
   end
 end
--- a/lib/cryptoexchange/client.rb
+++ b/lib/cryptoexchange/client.rb
@@ -49,7 +49,7 @@ module Cryptoexchange
           end
         end
       end
-      exchanges
+      exchanges.uniq.sort
     end
   end
 end
--- a/lib/cryptoexchange/client.rb
+++ b/lib/cryptoexchange/client.rb
@@ -8,6 +8,8 @@ module Cryptoexchange
     def pairs(exchange)
       pairs_classname = "Cryptoexchange::Exchanges::#{StringHelper.camelize(exchange)}::Services::Pairs"
       Object.const_get(pairs_classname).new.fetch
+    rescue HTTP::ConnectionError, HTTP::TimeoutError, JSON::ParserError
+      return { error: "#{exchange}'s service is temporarily unavailable." }
     end
 
     def ticker(market_pair)
@@ -24,6 +26,8 @@ module Cryptoexchange
           t.base == market_pair.base && t.target == market_pair.target
         end
       end
+    rescue HTTP::ConnectionError, HTTP::TimeoutError, JSON::ParserError
+      return { error: "#{exchange}'s service is temporarily unavailable." }
     end
 
     def available_exchanges
--- a/README.md
+++ b/README.md
@@ -84,6 +84,13 @@ Or install it yourself as:
   pairs = client.pairs('bitflyer')
 ```
 
+### List exchange services for certain currency
+```
+  client.exchange_for('btc')
+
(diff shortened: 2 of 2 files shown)


tmlee:
@dominicwong617 Maybe you can look into this

lib/cryptoexchange/client.rb
```diff
@@ -8,6 +8,8 @@ def initialize(ticker_ttl: 3)
     def pairs(exchange)
       pairs_classname = "Cryptoexchange::Exchanges::#{StringHelper.camelize(exchange)}::Services::Pairs"
       Object.const_get(pairs_classname).new.fetch
+    rescue HTTP::ConnectionError, HTTP::TimeoutError, JSON::ParserError
+      return { error: 'exchange service temporary unavailable' }
```
dominicwong617: @Swingcloud How about `#{exchange}'s service is temporarily unavailable.`

lib/cryptoexchange/client.rb
```diff
@@ -24,6 +26,41 @@ def ticker(market_pair)
...
+
+    def exchange_for(currency)
+      exchanges = {
+        base: [],
+        target: [],
+        temporary_unavailable: []
+      }
+      available_exchanges.each do |exchange|
+        pairs = pairs(exchange)
+        if pairs.is_a?(Hash)
```
dominicwong617: @Swingcloud it's more like `pairs.is_a?(Hash) && pairs[:error]` ?

lib/cryptoexchange/client.rb
```diff
@@ -24,6 +26,41 @@ def ticker(market_pair)
...
+          exchanges[:temporary_unavailable].push(exchange)
```
dominicwong617: `temporarily_unavailable` ?

lib/cryptoexchange/client.rb
```diff
@@ -24,6 +26,41 @@ def ticker(market_pair)
...
+          next
+        else
+          pairs(exchange).each do |pair|
```
dominicwong617: @Swingcloud why do you need it call it again?

lib/cryptoexchange/client.rb
```diff
@@ -24,6 +26,41 @@ def ticker(market_pair)
...
+            if pair.base == currency.upcase
+              exchanges[:base].push(exchange)
```
dominicwong617: @Swingcloud you need to make sure there are no duplicates

lib/cryptoexchange/client.rb
```diff
@@ -24,6 +26,41 @@ def ticker(market_pair)
...
+              break
+            elsif pair.target == currency.upcase
+              exchanges[:target].push(exchange)
+              break
+            end
+          end
+        end
+      end
+      exchanges
```
dominicwong617: @Swingcloud actually let's just return it as one array and ignore the ones that raises a HTTP error

Swingcloud: will do

lib/cryptoexchange/client.rb
Swingcloud: I call it again because some exchanges are at target, not at base

lib/cryptoexchange/client.rb
Swingcloud: will do

lib/cryptoexchange/client.rb
Swingcloud: my bad, I know what you mean

lib/cryptoexchange/client.rb
```diff
@@ -24,6 +26,30 @@ def ticker(market_pair)
...
+        folder_name.split('/').last
+      end
+    end
+
+    def exchange_for(currency)
+      exchanges = []
+      available_exchanges.each do |exchange|
+        pairs = pairs(exchange)
+        next if pairs.is_a?(Hash) && !pairs[:error].empty?
+        pairs.each do |pair|
```
dominicwong617: @Swingcloud you are still going to have duplicates, can you call `uniq` in the end?

dominicwong617: also, would be great if you can document this method in README

Swingcloud: seems like if it has an equal value on base or target, that will break and go to the other exchange.

and I'll document it

lib/cryptoexchange/client.rb
```diff
@@ -8,6 +8,8 @@ def initialize(ticker_ttl: 3)
     def pairs(exchange)
       pairs_classname = "Cryptoexchange::Exchanges::#{StringHelper.camelize(exchange)}::Services::Pairs"
       Object.const_get(pairs_classname).new.fetch
+    rescue HTTP::ConnectionError, HTTP::TimeoutError, JSON::ParserError
```
dominicwong617: @Swingcloud actually, where is `begin`?

Swingcloud: @dominicwong617  method "def" can serve as a "begin" statement
or we can put begin on it I guess, but it works the same 

https://stackoverflow.com/questions/1542672/how-does-one-use-rescue-in-ruby-without-the-begin-and-end-block

dominicwong617: @Swingcloud TIL, no worries

dominicwong617:
Let's try @Swingcloud @tmlee

PR #124, opened 2017-09-22
[#122] Implement Luno Ticker

tmlee:
Implement ticker for Luno exchange

https://github.com/coingecko/cryptoexchange/issues/122

- [/] I have added Specs
- [/] (If implementing Market Ticker) I have verified that the `volume` attributes is correct
- [/] (If implementing Market Ticker) I have verified that the `base` and `target` is correct

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ Or install it yourself as:
 | LakeBTC           | Y       |            |         |         | Y           |
 | Liqui             | Y       |            |         |         | Y           |
 | Livecoin          | Y       |            |         |         | Y           |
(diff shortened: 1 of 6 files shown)


dominicwong617:
@tmlee I think it supports individual ticker query
```
$ curl https://api.mybitx.com/api/1/ticker?pair=XBTZAR
```

tmlee:
@dominicwong617 Don't we want to prefer non-individual ticker api query to reduce hitting the API too often?

dominicwong617:
@tmlee yea you're right

PR #129, opened 2017-09-29
Mercury Cash API - Book and Returnticker

mercurycash:
Hello, this file has the method to request the books and the returnticker.

The ability to request the returnticker and book for transactions inside mercury cash
- [X] I have added Specs
- [X] (If implementing Market Ticker) I have verified that the `volume` attributes is correct
- [X] (If implementing Market Ticker) I have verified that the `base` and `target` is correct

### DIFF
--- a/mercurycash.js
+++ b/mercurycash.js
@@ -1,0 +1,20 @@
+var request = require('request');
+
+var Mercury = function(){};
+
+// Send an API request
+Mercury.bookethusd = function(callback){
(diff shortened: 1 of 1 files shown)


tmlee:
@mercurycash Our specs was unreliable earlier. If you rebase off from `master` now, specs should be passing. We are also a ruby library, if you are able to write the API wrapper in Ruby instead of JS following the existing structure that we currently have. Is MercuryCash an exchange or a brokerage service?

Issue #108, opened 2017-08-24
Implement Tux Exchange Ticker, Orderbook, and Trades


tmlee:
[X] Ticker https://github.com/coingecko/cryptoexchange/pull/130
[ ] Orderbook
[ ] Trades

Swingcloud:
I'll take this

PR #130, opened 2017-09-30
Tux exchange  pairs and ticker

Swingcloud:
- What is the purpose of this Pull Request? implement tux_exchange ticker and pairs
- What is the related issue for this Pull Request? #108 
- [x] I have added Specs


tmlee:
@Swingcloud You can rebase from master and try to get the specs to appear green. Its using VCR now

Swingcloud:
will do

PR #148, opened 2017-10-03
link to VCR switcheroo

rwhitworth:
The markdown for the link to the VCR project was flip-flopped.  So this PR puts it the right way around.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,6 +1,6 @@
 # Introduction
 
-### 
+###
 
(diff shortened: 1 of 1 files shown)


tmlee:
Thank you @rwhitworth , before i noticed this pull request, I went and move that section of the documentation to the wiki at https://github.com/coingecko/cryptoexchange/wiki/Implementing-a-New-Exchange#some-things-to-take-note
I implemented your change in that wiki, so i will close this for now

Issue #135, opened 2017-10-03
Implement Bit-Z Ticker, Orderbook, and Trades


tmlee:

Swingcloud:
I'll take this

PR #151, opened 2017-10-05
implement Bit-z pairs and ticker

Swingcloud:
- What is the purpose of this Pull Request? implement Bit-z pairs and ticker
- What is the related issue for this Pull Request? #135 
- [x] I have added Specs


tmlee:
@Swingcloud it is possible to implement ticker using single API call with https://www.bit-z.com/api_v1/tickerall ?

PR #163, opened 2017-10-10
Allcoin Ticker

yihyang:
  + Adding Ticker and OrderBook function for `Allcoin`
  + Wanted to add Trades but it seems that more work needs to be done on model before I can proceed
- [x] I have added Specs

### DIFF
--- a/spec/exchanges/allcoin/integration/market_spec.rb
+++ b/spec/exchanges/allcoin/integration/market_spec.rb
@@ -28,4 +28,21 @@ RSpec.describe 'Allcoin integration specs' do
     expect(DateTime.strptime(ticker.timestamp.to_s, '%s').year).to eq Date.today.year
     expect(ticker.payload).to_not be nil
   end
+
+  it 'fetch order book' do
+    btc_usd_pair = Cryptoexchange::Models::MarketPair.new(base: 'BTC', target: 'USD', market: 'allcoin')
+    order_book = client.order_book(btc_usd_pair)
+
+    expect(order_book.base).to eq 'BTC'
+    expect(order_book.target).to eq 'USD'
+    expect(order_book.market).to eq 'allcoin'
+    expect(order_book.asks).to_not be_empty
+    expect(order_book.bids).to_not be_empty
+    expect(order_book.asks.first.size).to eq 2
+    expect(order_book.bids.first.size).to eq 2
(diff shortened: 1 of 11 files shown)


spec/exchanges/allcoin/integration/market_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+
+    expect(ticker.base).to eq 'BTC'
+    expect(ticker.target).to eq 'USD'
+    expect(ticker.market).to eq 'allcoin'
+    expect(ticker.last).to be_a Numeric
+    expect(ticker.ask).to be_a Numeric
+    expect(ticker.bid).to be_a Numeric
+    expect(ticker.volume).to be_a Numeric
+    expect(ticker.timestamp).to be_a Numeric
+    expect(DateTime.strptime(ticker.timestamp.to_s, '%s').year).to eq Date.today.year
```
dominicwong617: @yihyang this will fail after 2017, right?

yihyang: @dominicwong617 just curious why will it fail?

the timestamp is retrieved from the following line:
https://github.com/yihyang/cryptoexchange/blob/b37763bab81cde0e6be6969c7e4b42e995311054/lib/cryptoexchange/exchanges/allcoin/services/market.rb#L34

Which is provided by the allcoin's API in here:
https://allcoin.com/About/APIReference/

dominicwong617: Because VCR locks the timestamp to be 2017, and Date.today.year will return 2018 once it is 2018?

tmlee: That's true and it may apply to some of the other specs too now that we are using VCR
How about checking if it falls within a reasonable range. Otherwise we get funny errors that we are not aware that we need to update VCR cassettes.

tmlee:
Thanks for your contribution @yihyang 
Will merge this; lets take care of the specs issue in a separate case https://github.com/coingecko/cryptoexchange/issues/174

spec/exchanges/allcoin/integration/market_spec.rb
yihyang: ok noted @tmlee @dominicwong617 thanks!

PR #164, opened 2017-10-10
Implement LiteBit.eu Ticker

weh:
  add LiteBit.eu ticker https://github.com/coingecko/cryptoexchange/issues/142
  https://github.com/coingecko/cryptoexchange/issues/142
- [X] I have added Specs
- [X] (If implementing Market Ticker) I have verified that the `volume` attributes is correct
- [X] (If implementing Market Ticker) I have verified that the `base` and `target` is correct

### DIFF
--- a/lib/cryptoexchange/exchanges/litebiteu/services/market.rb
+++ b/lib/cryptoexchange/exchanges/litebiteu/services/market.rb
@@ -39,6 +39,7 @@ module Cryptoexchange::Exchanges
           ticker.ask       = NumericHelper.to_d(market['buy'])
           ticker.bid       = NumericHelper.to_d(market['sell'])
           ticker.volume    = NumericHelper.to_d(market['volume'])
+          ticker.last      = (ticker.ask + ticker.bid) / 2
           ticker.timestamp = fetch_timestamp
           ticker.payload   = market
           ticker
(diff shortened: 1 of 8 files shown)


lib/cryptoexchange/exchanges/litebiteu/services/market.rb
```diff
@@ -0,0 +1,49 @@
...
+
+          ticker.base      = market_pair.base
+          ticker.target    = market_pair.target
+          ticker.market    = Litebiteu::Market::NAME
+          ticker.ask       = NumericHelper.to_d(market['buy'])
+          ticker.bid       = NumericHelper.to_d(market['sell'])
+          ticker.volume    = NumericHelper.to_d(market['volume'])
+          ticker.timestamp = fetch_timestamp
+          ticker.payload   = market
+          ticker
```
tmlee: I think `last` needs to be assigned. Aware that `last` attribute is not returned from the response which is weird.
We have an option to pick either the buy or sell or maybe take the average of the two?

weh: I added the average of ask/bid as `last` value.

tmlee:
👍  Thank you very much! @weh

Issue #169, opened 2017-10-14
Implement Poloniex Orderbook and Trades


dsantosmerino:
Working on it

PR #170, opened 2017-10-14
Feature/poloniex orders trades

dsantosmerino:
- What is the purpose of this Pull Request? Implement Poloniex orders and trades
- What is the related issue for this Pull Request? #169
- [x] I have added Specs


lib/cryptoexchange/exchanges/poloniex/services/order_book.rb
```diff
@@ -0,0 +1,40 @@
...
+          order_book.market    = Poloniex::Market::NAME
+          order_book.asks      = adapt_orders output['asks']
+          order_book.bids      = adapt_orders output['bids']
+          order_book.timestamp = Time.now.to_i
+          order_book.payload   = output
+          order_book
+        end
+
+        def adapt_orders(orders)
+          orders
```
dominicwong617: @dsantosmerino Could you make each element in the array to be

```
[price, amount, timestamp]
```

?

lib/cryptoexchange/exchanges/poloniex/services/order_book.rb
```diff
@@ -0,0 +1,40 @@
+require 'pry'
```
dominicwong617: @dsantosmerino Could you remove `pry`?

dsantosmerino: Oops, sorry!

dsantosmerino:
@dominicwong617 Same with this PR, I got this email notification: 'Could you make each element in the array to be `[price, amount, timestamp]`?', but nothing here.

PR #175, opened 2017-10-18
Add Order class to abstract formatting into [price, amount, timestamp]

tmlee:
When implementing OrderBook, there is no structure for individual Order.
Here is an abstraction of a Class, when assigned with price, amount, and timestamp will spit out the format we expect for all exchanges

### DIFF
--- a/lib/cryptoexchange/models/order.rb
+++ b/lib/cryptoexchange/models/order.rb
@@ -1,0 +1,13 @@
+module Cryptoexchange
+  module Models
+    class Order
+      attr_accessor :price, :amount, :timestamp
+
+      def initialize(args={})
+        @price      = args[:price]
+        @amount     = args[:amount]
+        @timestamp  = args[:timestamp] || nil
+      end
+    end
+  end
+end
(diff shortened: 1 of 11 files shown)


lib/cryptoexchange/models/order.rb
```diff
@@ -0,0 +1,17 @@
...
+    class Order
+      attr_accessor :price, :amount, :timestamp
+
+      def initialize(args={})
+        @price      = args[:price]
+        @amount     = args[:amount]
+        @timestamp  = args[:timestamp] || nil
+      end
+
+      def format_as_list
```
dominicwong617: @tmlee Actually, why do we need to do this? Can't we just return an array of `Cryptoexchange::Models::Order` ?

tmlee: @dominicwong617 That is possible. We just need to decide what end result we want to provide. Though i think returning as `Order` would give the user best flexibility how they want to store the results.

tmlee: @dominicwong617 now giving the user the choice to call this if they want to

dominicwong617: @tmlee let's just return an array of object as asks and bids, this choice doesn't add much benefits

Issue #174, opened 2017-10-18
Specs checking timestamp on a specific year will fail due to VCR capturing datetime


tmlee:
From the discussion at https://github.com/coingecko/cryptoexchange/pull/163/files#r144719383

Specs that are checking if timestamp assignment falls in the current year (reasonable date)
may fail when the year passes on.

Consider checking if it falls within a reasonable year

ztmeyer:
I'll pick this up

ztmeyer:
@tmlee how do you feel about using an `include` matcher? i.e.

```ruby
expect(1900..Date.today.year).to include(DateTime.strptime(ticker.timestamp.to_s, '%s').year)
```

Also it looks like there are a lot of these - I'll swap them all out.

tmlee:
@ztmeyer That approach seems fine, maybe date it back to `2000` or `1990`?
It shouldn't go any below than that. One of the purpose of this test is to ensure that the timestamp parsing makes sense; an erronous year that is out of range would likely mean that the implementation is incorrect.

ztmeyer:
Sounds good - I'll go for 2000, more to come

tmlee:
Many thanks @ztmeyer do issue a pull request with separate branches :shipit: :shipit:

ztmeyer:
Thanks @tmlee I created the PR - I switched over to using `Time.at` since it was more concise and had the same net effect. If you have any feedback let me know!

tmlee:
PR accepted. Thanks!

PR #177, opened 2017-10-20
Allow range for year match in tests

ztmeyer:
This resolves an issue with the test suite where the current year was being used as the reference in checking a timestamp. This would have caused test failures when the calendar year rolled over (in 2018).

https://github.com/coingecko/cryptoexchange/issues/174


tmlee:
👍  Thank you @ztmeyer

Issue #182, opened 2017-10-25
Implement Abucoins Ticker, Orderbook, Trades


tmlee:
Docs at https://docs.abucoins.com/

tmlee:
Implemented and merged

PR #187, opened 2017-10-27
Abucoins

akshaymankar:
   Solves #182 
   #182 
- [✓] I have added Specs
- [✓] (If implementing Market Ticker) I have verified that the `volume` attributes is correct
- [✓] (If implementing Market Ticker) I have verified that the `base` and `target` is correct


tmlee:
Looks good to me, thank you @akshaymankar 🎆

PR #190, opened 2017-10-30
[fixes #181] Add support for Exmo exchange

### LINKED ISSUE #181: Implement EXMO Ticker, Orderbook, Trades
Docs at https://exmo.com/en/api

falkoner:
Provide implementation for **Exmo** exchange API

#181 
- [x] I have added Specs


falkoner:
Got couple of questions. 

1. Volume. How to confirm correct value is used? Exmo provides 2:

> vol - the volume of deals within the last 24 hours
> vol_curr - the total value of all deals within the last 24 hours

So first one looks to be number of trades, and second is the value. Which one are we interested in? I've used **vol_curr** but suspect the currency used is in target, not base. Here is an example from their site:
```
{
  "BTC_USD": {
    "buy_price": "589.06",
    "sell_price": "592",
    "last_trade": "591.221",
    "high": "602.082",
    "low": "584.51011695",
    "avg": "591.14698808",
    "vol": "167.59763535",
    "vol_curr": "99095.17162071",
    "updated": 1470250973
  }
}
```

2.  `order_book` call supports optional parameter. Are we interested in using non-default?

> limit – the number of displayed positions (default: 100, max: 1000)
 
3.  So what's with rate limits / throttling? I don't see where this information is used or even stored for reference, but for Exmo it is as follows:

>  The number of API requests is limited to 180 per/minute from one IP address or by a single user.

tmlee:
Hi @falkoner thank you for your contribution
I should have highlighted it in the PR template as stated here https://github.com/coingecko/cryptoexchange/wiki/Implementing-a-New-Exchange#some-things-to-take-note

1) For a BTC/USD pair, the volume on our class should be in BTC (which is the base currency)
In that case, you will want to use `vol`

2) At this point, the library do not support custom display positions. I haven't seen it yet so far for the exchanges I've implemented; but maybe this is something that can reside in the exchange configuration overridable by the user per exchange.

3) Again at this point, we haven't thought about throttling yet, we pretty much give the power to the user to control. But I am open to suggestion if there is a nice way to handle and include this.

For (2) and (3), I am happy to create new issues so we can deal with them separately.

falkoner:
Ok, got it, @tmlee I've changed to `vol`  and resubmitted request. Will think about throttling and limit parameters.

tmlee:
@falkoner Thank you, PR merged! :shipit:

PR #192, opened 2017-11-01
[fixes #136] Add support for ACX exchange

### LINKED ISSUE #136: Implement ACX Ticker, Orderbook, and Trades

falkoner:
Implement support for ACX exchange

#136 

- [x] I have added Specs

### DIFF
--- a/lib/cryptoexchange/exchanges/acx/services/market.rb
+++ b/lib/cryptoexchange/exchanges/acx/services/market.rb
@@ -1,0 +1,51 @@
+module Cryptoexchange::Exchanges
+  module Acx
+    module Services
+      class Market < Cryptoexchange::Services::Market
+        class << self
+          def supports_individual_ticker_query?
+            false
+          end
+        end
+
+        def fetch
+          output = super ticker_url
+          adapt_all output
+        end
+
+        def ticker_url
+          "#{Cryptoexchange::Exchanges::Acx::Market::API_URL}/tickers"
+        end
+
+        def adapt_all(output)
+          output.map do |market_pair_key, ticker|
+            base, target = ticker["name"].split('/')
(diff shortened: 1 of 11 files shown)


falkoner:
Looks like there is bug on ACX side - trade type (called "side") returns null.

falkoner:
So what do you think, @tmlee ?

tmlee:
Sorry for the delay @falkoner 
In this case, we should not support Trade in this API at the moment until this is fixed or revisit this when somebody wants to use the trade feature for this exchange.

I'll look into what you have here and hopefully merge if all is good. thanks!

lib/cryptoexchange/exchanges/acx/services/market.rb
```diff
@@ -0,0 +1,45 @@
...
+
+        def adapt_all(output)
+          output.map do |market_pair_key, ticker|
+            adapt(ticker)
+          end
+        end
+
+        def adapt(output)
+          ticker           = Cryptoexchange::Models::Ticker.new
+          ticker.base      = output['base_unit']
```
tmlee: I would suggest just passing in `market_pair` and call `base` and `target` for assignment?
This will ensure the base/target are all uppercased

falkoner: Sure, done.

falkoner:
@tmlee as for trade, would you prefer to remove the class altogether from API for now?

tmlee:
@falkoner yeah i would drop support for that completely for this exchange

tmlee:
My apologies @falkoner I was reviewing the Pull Requests from top to bottom and merged another implementation of ACX exchange. 
I really wanted to include your proposal; will have to do a better job marking the issues as its being worked on in the Pull Requests.

falkoner:
np

PR #212, opened 2017-11-20
Implemented Coss Ticker

rusbal:
- Implementation of Coss Ticker
- No related issue
- Yes, I have added Specs
- The `volume` probably refers to BASE.  Please check.


rusbal:
Implemented CoinsMarkets Ticker.

tmlee:
Thanks! @rusbal @dominicwong617

tmlee:
Sorry @rusbal I looked at the PR again, I think there are some issues.
Maybe push as a hotfix since I've hit merged on this one.

For `coinsmarkets`
Example of the API
```
"BTC_LTC":{"id":1,"last":"0.008393939","lowestAsk":"0.009089880","highestBid":"0.008393944","24htrade":"1.885061167","high24hr":"1.500000000","low24hr":"0.000000999"}
```

In this case, our base should be LTC, target is BTC
That is because the price returned is in BTC term
And since the volume is in BTC and we want the volume to be in base (LTC), we have to divide against the last price

Similar issue as https://github.com/coingecko/cryptoexchange/pull/213

--------------------------------------

`COSS` looks good to me

PR #213, opened 2017-11-21
Implemented BCC Exchange Ticker

rusbal:
- Implemented BCC Exchange Ticker
- No related issue
- Yes, I have added Specs
- Probably `volume` refers to BASE

### DIFF
--- a/lib/cryptoexchange/exchanges/bitconnect/market.rb
+++ b/lib/cryptoexchange/exchanges/bitconnect/market.rb
@@ -1,0 +1,8 @@
+module Cryptoexchange::Exchanges
+  module Bitconnect
+    class Market
+      NAME = 'bitconnect'
+      API_URL = 'https://bitconnect.co/api'
+    end
+  end
+end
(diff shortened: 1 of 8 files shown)


tmlee:
@rusbal 
This is one example of the API that treats base and target differently
The API returned the following

```
{"status":"success !!","message":"BTC_BCC Market information","markets":[{"bid":"0.03604325","last_price":"0.03616990","volume24h":"3276.04684892","currency":"BCC","marketname":"BTC_BCC","ask":"0.03616980","low24h":"0.03029119","change24h":"-1.25%","high24h":"0.03684321","basecurrency":"BTC"}]}
```

I think the base/target is mapped incorrectly.
The base should be `BCC` and target should be `BTC`
If it were the other way round, that would mean 1 BTC = 0.036 BCC which is incorrect.
The volume seem to be in BTC according to https://bitconnect.co/user/trade?Market=BCC
So we probably want to divide the price to get volume in BCC which is our supposed base

tmlee:
duplicate commit in https://github.com/coingecko/cryptoexchange/pull/219

PR #214, opened 2017-11-21
Fix Base-Target position and volume

rusbal:
- Fix Base-Target position and volume
- No related issue for this Pull Request
- Yes, I have added Specs
- Yes, I have verified that the `volume` refers to BASE
- Yes, I have verified that the `base` and `target` is assigned correctly

PR #216, opened 2017-11-21
Feature/cex io trades

rusbal:
- Implementation of cex trades
- Issue 203
- Yes, I have added Specs
- Yes, I have verified that the `volume` refers to BASE


tmlee:
The base/target pair to be fixed here as well (from https://github.com/coingecko/cryptoexchange/pull/214/files/c909843ceccb16d73c1eba1080342e78cf58ff21..092e895da9438098c6f743b40d2c3f7ffdc810bc#r152446208)

lib/cryptoexchange/exchanges/cex/services/trades.rb
```diff
@@ -0,0 +1,32 @@
...
+  module Cex
+    module Services
+      class Trades < Cryptoexchange::Services::Market
+        def fetch(market_pair)
+          output = super(ticker_url(market_pair))
+          adapt(output, market_pair)
+        end
+
+        def ticker_url(market_pair)
+          "#{Cryptoexchange::Exchanges::Cex::Market::API_URL}/trade_history/#{market_pair.target}/#{market_pair.base}/?since=1"
```
tmlee: Do we need the `since=1`
It looks like that parameter is giving us old trades. Probably omit that if we want the latest current trades.

rusbal: `since=1` removed

Issue #121, opened 2017-09-22
Implement LocalBitcoins


tmlee:
https://localbitcoins.com/api-docs/

PR #235, opened 2017-11-30
implement Local bitcoins ticker, order book and trades

Swingcloud:
- What is the purpose of this Pull Request? implement local bitcoins ticker order book and trades 
- What is the related issue for this Pull Request? #121 
- [x] I have added Specs


lib/cryptoexchange/exchanges/local_bitcoins/services/order_book.rb
```diff
@@ -0,0 +1,44 @@
...
+          order_book.timestamp = Time.now.to_i
+          order_book.payload   = output
+          order_book
+        end
+
+        def adapt_orders(orders)
+          orders.collect do |order_entry|
+            price, amount = order_entry
+            Cryptoexchange::Models::Order.new(price: price,
+                                              amount: amount,
```
tmlee: @Swingcloud 
https://localbitcoins.com/bitcoincharts/sgd/orderbook.json
https://localbitcoins.com/api-docs/

The 2nd item, which is amount.
It looks like it refers to the maximum trade in the target currency
This is a little hairy and different than exchanges, think we should divide the amount with price.

It is different than usual exchange because someone is allowed to trade less than the orderbook amount

lib/cryptoexchange/exchanges/local_bitcoins/services/trades.rb
```diff
@@ -0,0 +1,33 @@
...
+        end
+
+        def adapt(output, market_pair)
+          output.collect do |trade|
+            tr = Cryptoexchange::Models::Trade.new
+            tr.trade_id  = trade['tid']
+            tr.base      = market_pair.base
+            tr.target    = market_pair.target
+            # there is no type in respond, I've send mail to ask them
+            # tr.type    = xxx
```
tmlee: I think for localbitcoins, its mainly connecting buyer seller as is.
Not really like an exchange where orderbook of maker and taker pushes the price up/down.
Should be ok to leave `nil` if they dont respond

lib/cryptoexchange/exchanges/local_bitcoins/services/order_book.rb
Swingcloud: sure let me look into this tomorrow

Swingcloud: @tmlee so we should just record price in this currency, and save that maximum amount somewhere?

Swingcloud:
I got their reply about trades, they said: 
```
Every trade is both a buy and sell, as someone buys and someone sells. As such we do not specify this directly.
```

so maybe I can leave type blank

Issue #138, opened 2017-10-03
Implement GetBTC Ticker, Orderbook, and Trades


tmlee:

Swingcloud:
I'll take this

Swingcloud:
https://getbtc.org/api-docs.php

PR #249, opened 2017-12-14
Implement GetBTC ticker, order book and trades 

Swingcloud:
- What is the related issue for this Pull Request? #138 
- [x] I have added Specs


Swingcloud:
@tmlee after I check all pairs ticker, suddenly found that volume is refer to all BTC trades volume, so I take it out

and the transactions volume I've fixed

lib/cryptoexchange/exchanges/getbtc/services/market.rb
```diff
@@ -28,7 +28,6 @@ def adapt(output, market_pair)
           ticker.last      = NumericHelper.to_d(ticker_data['last_price'])
           ticker.high      = NumericHelper.to_d(ticker_data['max'])
           ticker.low       = NumericHelper.to_d(ticker_data['min'])
-          ticker.volume    = NumericHelper.divide(NumericHelper.to_d(ticker_data['total_btc_traded']), ticker.last)
```
tmlee: @Swingcloud by removing this we wont have any information on volume, is there no way to convert it?

Swingcloud: seems no way, I've check this website, maybe we can do sum up all trades to get the single volume of currency, what do you think?

tmlee: We can use `total_btc_traded`

TJ-CoinGecko:
@tmlee @wongy91  can this proceed? Got another request from GetBTC guys

Issue #250, opened 2017-12-14
Implement Cryptobridge Ticker, Orderbook, and Trades


bobbyong:
https://crypto-bridge.org/

Swingcloud:
https://api.crypto-bridge.org/api/v1/ticker

Swingcloud:
I'll take this

PR #251, opened 2017-12-15
Implement CryptoBridge ticker

Swingcloud:
- What is the purpose of this Pull Request? as title, and there's only endpoint for ticker now
- What is the related issue for this Pull Request? #250 
- [x] I have added Specs

### DIFF
--- a/lib/cryptoexchange/exchanges/crypto_bridge/services/market.rb
+++ b/lib/cryptoexchange/exchanges/crypto_bridge/services/market.rb
@@ -1,0 +1,48 @@
+module Cryptoexchange::Exchanges
+  module CryptoBridge
+    module Services
+      class Market < Cryptoexchange::Services::Market
+        class << self
+          def supports_individual_ticker_query?
+            false
+          end
+        end
+
+        def fetch
+          output = super ticker_url
+          adapt_all(output)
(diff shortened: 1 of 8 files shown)


lib/cryptoexchange/exchanges/crypto_bridge/services/market.rb
```diff
@@ -0,0 +1,48 @@
...
+
+        def adapt(output, market_pair)
+          ticker = Cryptoexchange::Models::Ticker.new
+          ticker.base      = market_pair.base
+          ticker.target    = market_pair.target
+          ticker.market    = CryptoBridge::Market::NAME
+          ticker.ask       = NumericHelper.to_d(output['ask'])
+          ticker.bid       = NumericHelper.to_d(output['bid'])
+          ticker.last      = NumericHelper.to_d(output['last'])
+          ticker.volume    = NumericHelper.to_d(output['volume'])
```
tmlee: @Swingcloud Double check the volume, it should be in the base currency in which a divide (volume/last) is needed

Swingcloud: sure let me fix it!

PR #252, opened 2017-12-19
Lykke integration

estherleongym:
Added market.rb, pairs.rb, order_book.rb.
Added market_spec.rb, pairs_spec.rb
I have added Specs for Pairs, Ticker, and OrderBook
I have verified that the base and target is assigned correctly


lib/cryptoexchange/exchanges/lykke/services/pairs.rb
```diff
@@ -0,0 +1,54 @@
+module Cryptoexchange::Exchanges
+  module Lykke
+    module Services
+      class Pairs < Cryptoexchange::Services::Pairs
+        PAIRS_URL = "#{Cryptoexchange::Exchanges::Lykke::Market::API_URL}/AssetPairs/rate"
+        TARGETS = ['GBP', 'EUR', 'USD', 'CHF', 'ETH', 'BTC', 'JPY', 'LKK1Y', 'LKK2Y', 'LKK', 'ZAR', 'PLN', 'SEK', 'NOK', 'DILS', 'TRY',
```
Swingcloud: Hello @estherleongym seems like LYKKE api has endpoint that list all pairs they have in this url: 
https://public-api.lykke.com/api/AssetPairs/dictionary
I think if you use this and directly use `baseAssetId` and `quotingAssetId`
so we don't need to list target like this way

But I roughly checked there's some exception on output that
```
{
"id": "LKK2YCHF",
"name": "LKK2Y/CHF",
"accuracy": 5,
"invertedAccuracy": 5,
"baseAssetId": "91316bdc-f379-4442-98fd-f2c844995ca1",
"quotingAssetId": "CHF"
}
```
base id is wrong

```
{
"id": "GB032548USD",
"name": "GB032548",
"accuracy": 2,
"invertedAccuracy": 0,
"baseAssetId": "USD",
"quotingAssetId": "GB032548"
},
```
base id and quoting asset id seems swap, and market and some market required endpoint respond no content (maybe just don't put this in pair first)

and 

```
{
"id": "SHEELA",
"name": "SHEELA/EUR",
"accuracy": 2,
"invertedAccuracy": 6,
"baseAssetId": "SHEELA",
"quotingAssetId": "EUR"
},
```
you need to use id to call other market_pair required endpoint

tmlee: @Swingcloud @estherleongym 

Using this API seem to be a viable option.
I would just ignore all the bad pairs as stated above.

My thought process would be
1. Query https://public-api.lykke.com/api/AssetPairs/dictionary (set as the API url)
2. Adapt and assign base target directly

In the Ticker API
use the market pair base/target as the `id` under adapt
You would no longer need the strip functions and all, since the id will come from the market pair

If the marketpair is bad as @Swingcloud mentions, it will just be not found and ignored which is fine

lib/cryptoexchange/exchanges/lykke/services/market.rb
```diff
@@ -0,0 +1,74 @@
...
+        ticker           = Cryptoexchange::Models::Ticker.new
+        ticker.base      = market_pair.base
+        ticker.target    = market_pair.target
+        ticker.market    = Lykke::Market::NAME
+        ticker.bid       = NumericHelper.to_d(pair_output['bid'])
+        ticker.ask       = NumericHelper.to_d(pair_output['ask'])
+        ticker.last      = NumericHelper.to_d(pair_output['lastPrice'])
+        ticker.volume    = NumericHelper.to_d(pair_output['volume24H'])
+        ticker.timestamp = Time.now.to_i
+        ticker.payload   = raw_response
```
tmlee: The payload is the raw response for the Ticker itself rather than the entire response.
You can assign the pair_output to `ticker.payload`, you won't need the `raw_response` here and also under `Market.rb`
@estherleongym

lib/cryptoexchange/services/market.rb
```diff
@@ -1,6 +1,7 @@
 module Cryptoexchange
   module Services
     class Market
+      attr_reader :raw_response
```
tmlee: `raw_response` can be removed after https://github.com/coingecko/cryptoexchange/pull/252/files#r158582042

lib/cryptoexchange/client.rb
```diff
@@ -1,6 +1,6 @@
 module Cryptoexchange
   class Client
-
+
```
tmlee: just nitpicking, can remove this line

lib/cryptoexchange/exchanges/lykke/services/pairs.rb
```diff
@@ -0,0 +1,34 @@
+module Cryptoexchange::Exchanges
+  module Lykke
+    module Services
+      class Pairs < Cryptoexchange::Services::Pairs
+        PAIRS_URL = "#{Cryptoexchange::Exchanges::Lykke::Market::API_URL}/AssetPairs/rate"
```
tmlee: This is no longer in use it seems, if so can be removed

lib/cryptoexchange/exchanges/lykke/services/market.rb
```diff
@@ -0,0 +1,58 @@
...
+      class Market < Cryptoexchange::Services::Market
+        class << self
+          def supports_individual_ticker_query?
+            false
+          end
+        end
+
+        def fetch
+          output = super(ticker_url)
+          pairs_output = HTTP.get(dictionary_url)
```
tmlee: Instead of making 2 queries, an additional the the dictionary_url (since its already done in the Pairs side)
Would it be possible to just directly/simply parse the ticker_url only.

Then when a pair is passed to find the Ticker, if its something it does not understand because it came from dictionary with a weird different base/target, it will just fail and return nothing

tmlee:
@estherleongym looks good to me now, merged thanks!

PR #263, opened 2018-01-06
Implement Cryptopia order book and trades

pbalaban:
- What is the purpose of this Pull Request? _as title_
- What is the related issue for this Pull Request? _none_
- [x] I have added Specs


tmlee:
Many thanks for your contribution!

PR #264, opened 2018-01-06
implemented cobinhood pairs, markets, orderbooks

mwlang:
- Introduces Cobinhood exchange with implementation of Market, Ticker, and Order book.
- [Y] I have added Specs
- [N] (If implementing Market Ticker) I have verified that the `volume` refers to BASE
- [Y] (If implementing Market Ticker) I have verified that the `base` and `target` is assigned correctly

I am not 100% sure how to confirm volume is indeed expressed in base.  However, I did confirm the Volume matches what's displayed on Cobinhood's exchange website.

### DIFF
--- a/lib/cryptoexchange/exchanges/cobinhood/services/market.rb
+++ b/lib/cryptoexchange/exchanges/cobinhood/services/market.rb
@@ -1,0 +1,49 @@
+module Cryptoexchange::Exchanges
+  module Cobinhood
+    module Services
+      class Market < Cryptoexchange::Services::Market
+        class << self
+          def supports_individual_ticker_query?
+            true
+          end
+        end
+
+        def fetch(market_pair)
+          output = super(ticker_url(market_pair))
+          adapt(output, market_pair)
+        end
+
+        def ticker_url(market_pair)
+          trading_pair_id = "#{market_pair.base}-#{market_pair.target}"
+          "#{Cryptoexchange::Exchanges::Cobinhood::Market::API_URL}/market/tickers/#{trading_pair_id}"
+        end
+
+        def adapt(output, market_pair)
+          ticker = Cryptoexchange::Models::Ticker.new
+          market = output['result']["ticker"]
+
+          ticker.base      = market_pair.base
+          ticker.target    = market_pair.target
+          ticker.market    = Cobinhood::Market::NAME
+          ticker.last      = NumericHelper.to_d(market['last_trade_price'])
+          ticker.high      = NumericHelper.to_d(market['24h_high'])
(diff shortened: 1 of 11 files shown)


mwlang:
The CI failed with system errors where DNS for rubygems.org was not being resolved so nothing could be built.  If the build can be re-triggered, it should pass all specs.

lib/cryptoexchange/exchanges/cobinhood/services/pairs.rb
```diff
@@ -0,0 +1,31 @@
+module Cryptoexchange::Exchanges
+  module Cobinhood
+    module Services
+      class Pairs < Cryptoexchange::Services::Pairs
+        PAIRS_URL = "#{Cryptoexchange::Exchanges::Cobinhood::Market::API_URL}/market/trading_pairs"
+
+        def fetch_via_api(endpoint = self.class::PAIRS_URL)
+          ctx = OpenSSL::SSL::SSLContext.new
+          ctx.verify_mode = OpenSSL::SSL::VERIFY_NONE
```
tmlee: Is it possible to make do without SSL verify none?

lib/cryptoexchange/exchanges/cobinhood/services/market.rb
```diff
@@ -0,0 +1,49 @@
...
+
+          ticker.base      = market_pair.base
+          ticker.target    = market_pair.target
+          ticker.market    = Cobinhood::Market::NAME
+          ticker.last      = NumericHelper.to_d(market['last_trade_price'])
+          ticker.high      = NumericHelper.to_d(market['24h_high'])
+          ticker.low       = NumericHelper.to_d(market['24h_low'])
+          ticker.ask       = NumericHelper.to_d(market['lowest_ask'])
+          ticker.bid       = NumericHelper.to_d(market['highest_bid'])
+          ticker.volume    = NumericHelper.to_d(market['24h_volume'])
```
tmlee: 👍

lib/cryptoexchange/exchanges/cobinhood/services/pairs.rb
mwlang: Right now, Cobinhood's SSL certificate is not valid.  The only way I found around it was to not verify it.  Perhaps merge for now, but keep an open issue ticket to check every 30 days or so so see if they've fixed their certificate issues?

tmlee: @mwlang Tested on my local machine without SSL seem to work fine, i'll accept this PR and remove the verify none SSL and see how this goes moving forward

tmlee:
:+1: Thank you!

PR #280, opened 2018-01-15
Ethfinex integration

estherleongym:
- Implemented Pairs, Tickers, and OrderBook
- [X] I have added Specs


Swingcloud:
Hello @estherleongym seems like your master not yet in the latest version, maybe it's cause the test failed, would you rebase the master then we can see the test pass or not?

Issue #275, opened 2018-01-09
Implement Coinrail Ticker, Orderbook, Trades


tmlee:
https://coinrail.co.kr/api/document

Looks like there is no API to fetch pairs, consider using YML if no choice

estherleongym:
Working on this!

PR #311, opened 2018-02-01
Coinrail integration

estherleongym:
- Implemented Pairs, Tickers, OrderBook, Trades and added to README for Coinrail
- Related to #275 
- [ X ] I have added Specs


lib/cryptoexchange/exchanges/coinrail/services/market.rb
```diff
@@ -0,0 +1,56 @@
...
+                output.push(pair)
+              }
+            end
+          end
+          adapt_all(output)
+        end
+
+        def ticker_url
+          # No API endpoint available to obtain tickers (as at 29/01/2018), hence the URL below:
+          "https://coinrail.co.kr/main/market_info?v=#{(Time.now.to_f * 1000).to_i}"
```
Swingcloud: great you find this url that state all pairs and ticker, so maybe we can use this to list app pairs and not to use `yml` file

btw is it necessary to add params v ? cause it will make test fail (vcr can not handle url changing)

Swingcloud: and you can use this url to list pairs like:
grab 
```
"cs_name": "BTC",
"ce_name": "ETH",
```
and it's done!

Swingcloud: and if you use this url in paris.rb, you can extract it out to `exchanges/coinrail/market.rb` and give it another name, so we can father all url in one place

estherleongym: Ah OK, will amend and push the revision up :)

estherleongym: @Swingcloud added in commit aee6cdf!

Swingcloud:
great job!

Swingcloud:
@tmlee I think this is good to merge!

Issue #312, opened 2018-02-02
Implement Big.One Ticker, Orderbook, Trades


tmlee:
Docs at https://developer.big.one/

PR #313, opened 2018-02-03
BigONE Integration

estherleongym:
- Implemented Pairs, Tickers, OrderBook, Trades and added to README for BigONE
- Related to Issue #312 
- [ X ] I have added Specs


lib/cryptoexchange/exchanges/bigone/market.rb
```diff
@@ -0,0 +1,8 @@
+module Cryptoexchange::Exchanges
+  module Bigone
+    class Market
+      NAME = 'bigone'
+      API_URL = 'https://api.big.one/markets'
```
tmlee: Best to have API_URL be `https://api.big.one` as markets may not been appended in all endpoints especially if we implement User Auth in the future

lib/cryptoexchange/exchanges/bigone/services/market.rb
```diff
@@ -0,0 +1,36 @@
+module Cryptoexchange::Exchanges
+  module Bigone
+    module Services
+      class Market < Cryptoexchange::Services::Market
+        class << self
+          def supports_individual_ticker_query?
+            true
```
tmlee: Since it is possible to get all the tickers using `https://api.big.one/markets`, this would be preferred vs making multiple individual API calls. Consider implementing `supports_individual_ticker_query?` as false and the corresponding adapt for a single API call.

lib/cryptoexchange/exchanges/bigone/services/pairs.rb
```diff
@@ -0,0 +1,23 @@
...
+  module Bigone
+    module Services
+      class Pairs < Cryptoexchange::Services::Pairs
+        PAIRS_URL = "#{Cryptoexchange::Exchanges::Bigone::Market::API_URL}"
+
+        def fetch
+          output = super
+          market_pairs = []
+          output.each do |pair|
+            market_pairs << Cryptoexchange::Models::MarketPair.new(
```
tmlee: looking at the API response, I think you should be able to access the base/quote using the following

```
output['data'].each do |pair|
  pair['base']
  pair['quote']
end
```

estherleongym:
@tmlee OK done!

Issue #319, opened 2018-02-07
use Ethplorer API to recognise token symbol via address


Swingcloud:
https://github.com/EverexIO/Ethplorer/wiki/Ethplorer-API 

this service can let use use their api to recognise some token just have address on it ( like EtherDelta)

but somehow it took a lot of time because once it face nonstandard token, it will go to that api and nearly take almost one second to get back symbol

so theoretically we can use this, but respond about that exchange will be very very slow

PR #320, opened 2018-02-07
Implement Token helper

Swingcloud:
- What is the purpose of this Pull Request? use token helper to recognise nonstandard token
- What is the related issue for this Pull Request? #319


lib/cryptoexchange/helpers/token_helper.rb
```diff
@@ -0,0 +1,22 @@
+module TokenHelper
+  class << self
+    API_URL = 'https://api.ethplorer.io/getTokenInfo'
+    def get_symbol(token)
+      endpoint = "#{API_URL}/#{token}?apiKey=buoc5056EEZGf105"
```
tmlee: @Swingcloud Not too good having an API key dropped in here, we should make it configurable from the user end. Consider using a config file

Issue #298, opened 2018-01-24
Implement Right BTC Decentralized Exchange Ticker, Orderbook, Trades 


bobbyong:
https://www.rightbtc.com/#!/home

Swingcloud:
can't find any api doc on website, so I write an email to ask it

Swingcloud:
https://www.rightbtc.com/api/trader/

estherleongym:
Working on this!

PR #323, opened 2018-02-10
RightBTC Integration

estherleongym:
- Implemented Pairs, Tickers, OrderBook, Trades and added to README for RightBTC
- Related to Issue #298 
- [X] I have added Specs


lib/cryptoexchange/exchanges/rightbtc/services/market.rb
```diff
@@ -0,0 +1,66 @@
...
+          ticker.high      = NumericHelper.to_d(output['high24h'])
+          ticker.low       = NumericHelper.to_d(output['low24h'])
+          ticker.volume    = NumericHelper.to_d(output['vol24h'].abs)
+          ticker.timestamp = Time.now.to_i
+          ticker.payload   = output
+          ticker
+        end
+
+        def strip_pairs(pair)
+          #Match last 3 or 4 if it hits target
```
tmlee: @estherleongym What do you think of this approach?
Make additional call to `pairs` api to get the list of possible unique targets rather than defining it in a constant.
Extra call yes, but at least we are making only 2 calls and dont have to worry about the target currencies?

estherleongym: @tmlee okay agreed; revised it to call the `pairs` api and to get a match from the `tickers` output.

PR #341, opened 2018-03-02
Add lbank exchange

wenjieyek:
- [ ] I have added Specs

added lbank market list and ticker to the gem

### DIFF
--- a/lib/cryptoexchange/exchanges/lbank/services/market.rb
+++ b/lib/cryptoexchange/exchanges/lbank/services/market.rb
@@ -8,43 +8,40 @@ module Cryptoexchange::Exchanges
           end
         end
 
-        def fetch
-          output = super(ticker_url)
-          adapt_all(output)
+        def fetch(market_pair)
+          output = super(ticker_url(market_pair))
+          adapt(output, market_pair)
         end
 
-        def ticker_url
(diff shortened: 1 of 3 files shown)


lib/cryptoexchange/exchanges/lbank/services/market.rb
```diff
@@ -0,0 +1,46 @@
...
+          ticker.payload   = output
+          ticker
+
+        end
+
+         
+        end
+      end
+    end
+  end
```
tmlee: Indentation can be fixed

lib/cryptoexchange/exchanges/lbank/services/trades.rb
```diff
@@ -0,0 +1,35 @@
...
+            tr = Cryptoexchange::Models::Trade.new
+            tr.base      = market_pair.base
+            tr.target    = market_pair.target
+            tr.market    = Lbank::Market::NAME
+
+            tr.trade_id  = trade['tid']
+            tr.type      = trade['type']
+            tr.price     = trade['price']
+            tr.amount    = trade['amount']
+            tr.timestamp = trade['date_ms']
```
tmlee: @wenjieyek The timestamp is also in `ms` here. Consider writing a spec to test this case and make it pass when converting to seconds?

PR #352, opened 2018-03-11
Integrate trade data from bit2c crypto exchange

avinoamOltchik:
- Please add trade data from Bit2c.co.il, it is a cryptocurrency exchange that enables trade in Bitcoin since March 2013.
- Our API is available on : 
https://bit2c.co.il/home/api

I'm available for any questions or inquiries at avinoam@bit2c.co.il


tmlee:
@avinoamOltchik please create an Issue if there is a request to support a new API of an exchange
Pull Request are for proposal, use this if your development team can propose an implementation to be merged into this gem.

PR #360, opened 2018-03-12
Feature/coinroom

cySheng:
- Implemented Pairs, Tickers, OrderBook, Trades and updated README for CoinFalcon
- Related to issue #337
- [ X ] I have added Specs
- [ X ] (If implementing Market Ticker) I have verified that the volume refers to BASE
- [ X ] (If implementing Market Ticker) I have verified that the base and target is assigned correctly

Issue #337, opened 2018-02-27
Implement CoinRoom Ticker, Orderbook, and Trades


tmlee:
https://coinroom.com/public-api

cySheng:
Working on this

PR #367, opened 2018-03-14
Feature/coinroom ticker only

cySheng:
- Implemented Pairs, Ticker & updated Readme for coinroom 
- Related to issue #337 
- [x] I have added Specs

Coinroom has orderbook and market api but it wasn't included in this branch. Refer to #360.


lib/cryptoexchange/exchanges/coinroom/coinroom.yml
```diff
@@ -0,0 +1,97 @@
...
+  - :base: ZCOIN
+    :target: GBP
+  - :base: LSK
+    :target: GBP
+  - :base: ZEC
+    :target: GBP
+  - :base: BURST
+    :target: GBP
+  - :base: VIA
+    :target: GBP
```
wongy91: the api returns us
https://coinroom.com/api/availableCurrencies

```
{
    "real": [
        "USD",
        "EUR",
        "PLN",
        "GBP",
        "NOK",
        "CHF",
        "CZK",
        "DKK"
    ],
    "crypto": [
        "BTC",
        "LTC",
        "ETH",
        "DASH",
        "XMR",
        "BCC",
        "PIVX",
        "ZCOIN",
        "LSK",
        "ZEC",
        "BURST",
        "VIA"
    ]
}
```

basically this list is made up of combinations of results["real"] X results["crypto"].

are we able to have the logic (to populate the combinations) instead of fetching from this file? 1 less thing to maintain.

cySheng: Yup, what you're suggesting is possible. The gist of the logic is basically in the methods (find_and_write_all_possible_pairs) in pairs.rb. However, instead of returning the list of pairs using this logic, I've chosen to write the pairs to the YML file instead. 

We can have it populate the combinations instead of fetching form the file but there is a couple of problems. 
There is 8 real currencies and 12 crypto currencies. The total combinations is 96. However, some of those combinations are not possible. For example, an api request with any crypto currency paired with real currency ("DKK") will return no values:
```
{
  "low": 0,
  "high": 0,
  "vwap": 0,
  "volume": 0,
  "last": 0,
  "ask": 0,
  "bid": 0
}
```

So the choice now is whether to include all the pairs regardless of whether it returns valid values or not or we only include the pairs that only have valid values. 
1) If we choose to go for the first option of including all pairs regardless, out of 96 possibilities, **half** of it (~48) will return multiple values of 0 as per above. 
If you refer to the YML file, it basically contains all the _VALID_ possible combinations of different pairs. 
2) If we check whether the value are valid, there are two possible options: 
a) Don't use the YML file - During the request for the pairs of coinroom, you send an api request to check whether the different combination of pairs to return valid values. If yes, insert it into the array of pairs. The only thing that is not feasible with this method I believe is you need to query the api for all the list of possible combinations which will make it very slow and is not practical.
b) Use the YML file - Only write the VALID pairs into the yml file. (The current method)

To be honest, I looked at all the options and I went with 2b. The other possible option is no. 1 but half of the pairs returned will have no information/data about it when using it to query tickers. If you're fine with that, I could go ahead with that.
So far these are all the possibilities I can think of, if there is any more please do share.

wongy91: Thanks for the explanation 👍I agree with you that we stick to the YML for now.

Also do consider discussion below:
```
{
  "data": [],
  "result": true
}
```

- The data returned for DKK/LTC pair is empty, however it is valid response (HTTP Code 200, and `response['result']` is `true`). 

- So the question here is should we filter the response returned by exchange (API) or should the gem acts like a blackbox so it does its job if the API says a response is valid?

cySheng: If I understand you correctly, I believe that we should filter the response returned by the exchange (API). I think it would be annoying for users to think that a combination such as DKK/LTC is possible but when it queries the tradebook or ticker with it, there is no data. 
So I believe we should filter the response. 
Correct me if i'm wrong, if we stick to our current method, we are basically filtering the response right?

tmlee: 1. I think we cannot depend on the response code. These exchanges, sometimes return 200 and inserts the error code in the result. Although some exchange do uses the code accordingly, but not this one.

2. Looking at https://coinroom.com/stock there seem to be no instance where DKK is base and LTC is target giving DKK/LTC, so we do not need to worry about the reverse. Hence we continue to only permute REAL/CRYPTO

3. I like to use the pair API because it makes the pair management transparent. Less pairs to manually manage the better it is. Hence, if we know that all zeroes are an invalid response. 
You could explore something like this https://github.com/coingecko/cryptoexchange/blob/master/lib/cryptoexchange/exchanges/bittrex/services/market.rb#L42

4. Best of ability, the gem should work like a blackbox. Get a list of pairs (from API) and be able to call tickers, etc on it continuously.

@wongy91 @cySheng  thoughts?

cySheng: So you're saying we just include all the pairs regardless (not using the yml file)? If a user request a ticker using a non valid cryptocurrency and real currency pair (for e.g. DKK/LTC), we should be raising a `  Cryptoexchange::ResultParseError `? 

I mean if we want it to act like a blackbox then I don't think we have a choice. Some users might find it annoying when certain calling the ticker api through certain pairs and this error arises but it might be the only choice in this situation. 

If @wongy91 think it's alright, then we should just proceed with this blackbox approach.

tmlee: @cySheng yes. in this case, DKK/LTC is quite not likely. LTC/DKK is more likely. That is `Crypto/Real` based on the json above. 

At the moment using the gem, we are already getting a bunch of HTTP errors and what not which we have to handle, not to mention rate limiting and some erronous results. The exception should be accounted for.

Yea, hardly the yml can be easily maintained unless informed about it

wongy91: agree that we keep the Crypto/Real(Fiat) valid.

cySheng:
Hi guys, as per instructed, I have removed the usage of YML file and am just using all the available currency pairs from the API. Do have a look!

tmlee:
Looks good to me, easier with the pairs pulled via API

Issue #359, opened 2018-03-12
Append Orderbook, Trades to Coinone


wongy91:
order_book - https://doc.coinone.co.kr/#api-Public-Orderbook
trades - https://doc.coinone.co.kr/#api-Public-RecentTransactions

PR #371, opened 2018-03-14
append coinone (existing) - orderbook and trades

wongy91:
  - append order_book
  - append trades
  - refactor `market.rb` to be relying on `market_pair`'s attribute instead of hardcoding. (single source of truth)

https://github.com/coingecko/cryptoexchange/issues/359

- [x] I have added Specs


tmlee:
👍 Nice!

PR #373, opened 2018-03-15
Add NLexch.com Exchange

MassICT:
       Add new exchange
- [x ] I have added Specs
- [x ] (If implementing Market Ticker) I have verified that the `volume` refers to BASE


tmlee:
Thanks for the contribution @MassICT

Issue #176, opened 2017-10-19
Implement BitMEX Ticker, Orderbook, and Trades


tmlee:

yihyang:
@tmlee I will take this up

Swingcloud:
https://www.bitmex.com/app/restAPI

estherleongym:
working on this!

PR #376, opened 2018-03-17
BitMEX Integration

estherleongym:
- Implemented Pairs, Tickers, OrderBook, Trades and updated README
- Related to issue #176
- [ X ] I have added Specs


lib/cryptoexchange/exchanges/bitmex/services/order_book.rb
```diff
@@ -0,0 +1,55 @@
...
+          {asks: asks, bids: bids}
+        end
+
+        def adapt(output, market_pair)
+          order_book = Cryptoexchange::Models::OrderBook.new
+          timestamp = Time.now.to_i
+
+          order_book.base      = market_pair.base
+          order_book.target    = market_pair.target
+          order_book.market    = Bitmex::Market::NAME
```
wongy91: better to replace with `market_pair.market` for 1 less dependency?

Issue #383, opened 2018-03-26
Implement Coinue Ticker, Orderbook, Trades


bobbyong:
https://coinut.com/
API: https://github.com/coinut/api/wiki

cySheng:
I'll do this

wisqo:
Need any help on this?

tmlee:
@cySheng How to make calls to coinut

1. Register an account
2. Get API key and keep that safe somewhere
3. Take note your username
4. Get POSTMAN app to test calls https://www.getpostman.com/
5. Now lets make the call, but before that, we need to sign the request with our api key, do it with the following
```
require 'openssl'
hmac_hex =  (OpenSSL::HMAC.hexdigest(OpenSSL::Digest.new('sha256'),  <API_KEY>, <PAYLOAD>)).strip
```
6. Payload is the API we want to call, refer to https://github.com/coinut/api/wiki/Websocket-API
For example we want to get Ticker data so lets refer to https://github.com/coinut/api/wiki/Websocket-API#get-realtime-ticks
Copy 
```
{"request":"inst_tick","inst_id":1, "subscribe": true, "nonce":615146544} and put it into hmac payload above
```
6. Copy the hmac hex, on POSTMAN create the following request
```
POST request
URL: https://api.coinut.com/
body: {"request":"inst_tick","inst_id":1, "subscribe": true, "nonce":615146544}
headers:
X-USER: <username>
X-SIGNATURE: <hmac_hex>
```
Note: body must be the same as the payload we used when signing hmac

7. Execute query

PR #417, opened 2018-04-08
Feature/coinut

cySheng:
The purpose of this pull request is to integrate coinut api.
Issue #383 
- [ ] I have added Specs

This pull request is still a WIP.
1) All request to this api requires the user to be authenticated. 
In order for users to user the gem, a file called 

> coinut_auth.yml

 is required to be in their config/cryptoexchange folder.

I am not sure whether this is the best way to go about things so please do suggest alternatives. 

2) Several methods that were needed in services/pairs.rb is now needed in the market, orderbook and trades as to authenticate with the api. For example, path methods such as `auth_details_path` and `exchange_class`. Other methods that are repeated are `generate_nonce`, `generate_payload`, etc. 

Where exactly is the best place to call and include the methods to be reused?
Firstly, I was going to include helpers into the helpers folder but it seems that all the helper method there are reserved adding methods to BUILT-IN ruby classes.
Thoughts?

3) I am also not sure how to get started on the SPECs as it would seem that each request to the api would need a username and an api key.


tmlee:
@cySheng 

1. coinut_auth works.
Other alternative, if i were you would be to generalize it as `cryptoexchange_api_keys.yml`
```
coinut:
  key:...
  username: ...
```
which will support multiple exchanges.

Another option is to use environment variable, but that will pollute if we start implementing it for other exchanges. So this approach is fine


2. Since this is specific to Coinut we can include a helper class under the coinut directory which can referred by these classes

3. Maybe we can use a test username/key to make the call. Then capture it via VCR.
If we cant get a new username/key just make a call with your key and change the VCR file to remove your personal credentials. That suffice for a spec.

cySheng:
1. I have changed `coinut_auth.yml` to be `cryptoexchange_api_keys.yml` instead. Definitely seems like a better option for the future if there is any similar APIs out there.

I have also added a pending method to the rspec test to ensure travis ci pass. Let me know if anything needs to be changed.

cySheng:
Through some bypass, I am also able to generate the cassettes for orderbook and trades. 
However, there is something wrong with the rspec test for both order book and trades. 

For, order_book the output should be something like this:
```
{"buy"=>
  [{"count"=>2, "price"=>"10748.18", "qty"=>"0.03700000"},
   {"count"=>1, "price"=>"10732.33", "qty"=>"0.00883000"},
   {"count"=>1, "price"=>"10694.94", "qty"=>"0.00261000"},
```

However, it emulates the previous request which is the ticker call instead, returning this: 

```
{"buy"=>
  [{"count"=>2, "price"=>"10748.18", "qty"=>"0.03700000"},
   {"count"=>1, "price"=>"10732.33", "qty"=>"0.00883000"},
   {"count"=>1, "price"=>"10694.94", "qty"=>"0.00261000"},
```

I suspect that its due to the nonce that is generated and added into the payload/headers.

Will have a deeper look into it when time allows it.

PR #446, opened 2018-04-18
Add FreiExchange ticker and orderbook

wenjieyek:
- [Y] I have added Specs


lib/cryptoexchange/exchanges/freiexchange/services/pairs.rb
```diff
@@ -0,0 +1,23 @@
+module Cryptoexchange::Exchanges
+  module Freiexchange
+    module Services
+      class Pairs < Cryptoexchange::Services::Pairs
+
+        def fetch
+          output = super
+          market_pairs = []
+          output.each do |pair|
+            market_pairs << Cryptoexchange::Models::MarketPair.new(
```
tmlee: Instead of using the yml list, you can use https://freiexchange.com/api/returnTicker to return the list of pairs
In fact you can also use https://freiexchange.com/api/returnTicker for the ticker. In that way we make 1 call instead of individual.

The volume is in btc, so we must also multiply it by price to get the volume by base

@wenjieyek

Issue #400, opened 2018-04-04
Implement Graviex Ticker, Orderbook, Trades


TJ-CoinGecko:
API: https://graviex.net/documents/api_v2?lang=en
API Return All Tickets with Price, Volume, High, Low, Etc: https://graviex.net:443//api/v2/tickers.json
Return Specifc Market: https://graviex.net:443//api/v2/tickers/{market}.json

Website: https://graviex.net/
Logo: https://graviex.net/assets/logo-b9cd2fc6aa72e4117f4b7f70c1e5278f.png
Twitter: https://twitter.com/graviex_net

PR #421, opened 2018-04-12
Add Mercado Niobio Cash, TemBTC and Graviex

vitorgamer58:
- What is the purpose of this Pull Request? Add Mercado Niobio Cash, Graviex and TemBTC
- What is the related issue for this Pull Request? https://github.com/coingecko/cryptoexchange/issues/420 , https://github.com/coingecko/cryptoexchange/issues/402 and https://github.com/coingecko/cryptoexchange/issues/400
- [ ] I have added Specs
- [ ] (If implementing Market Ticker) I have verified that the `volume` refers to BASE
- [X] (If implementing Market Ticker) I have verified that the `base` and `target` is assigned correctly

PR #453, opened 2018-04-20
GRAVIEX added

demuskov:
- GRAVIEX integration
- https://github.com/coingecko/cryptoexchange/issues/400
- [X] I have added Specs

### DIFF
--- a/lib/cryptoexchange/exchanges/graviex/graviex.yml
+++ b/lib/cryptoexchange/exchanges/graviex/graviex.yml
@@ -1,0 +1,68 @@
+:pairs:
+  - :base: GIO
+    :target: BTC
+  - :base: GIO
+    :target: DOGE
+  - :base: GIO
+    :target: LTC
+  - :base: GIO
+    :target: ETH
+  - :base: PSD
+    :target: BTC
+  - :base: PHC
+    :target: BTC
+  - :base: XGC
+    :target: BTC
+  - :base: DEV
+    :target: BTC
+  - :base: PBS
+    :target: BTC
+  - :base: DIN
+    :target: BTC
+  - :base: ADV
+    :target: BTC
+  - :base: DV7
+    :target: BTC
+  - :base: ARGO
+    :target: BTC
+  - :base: ESCO
+    :target: BTC
+  - :base: NEET
+    :target: BTC
+  - :base: XYLO
+    :target: BTC
+  - :base: STEEP
+    :target: BTC
+  - :base: CRFT
+    :target: BTC
+  - :base: ENV
+    :target: BTC
+  - :base: SNG
+    :target: BTC
+  - :base: NYC
+    :target: BTC
+  - :base: ZOC
+    :target: BTC
+  - :base: BTCZ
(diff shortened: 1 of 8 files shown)


vitorgamer58:
good

lib/cryptoexchange/exchanges/graviex/graviex.yml
```diff
@@ -0,0 +1,68 @@
+:pairs:
```
tmlee: Since they have an API returning list https://graviex.net/api/v2/markets.json
That would be preferred over manual yml @demuskov

lib/cryptoexchange/exchanges/graviex/services/market.rb
```diff
@@ -0,0 +1,84 @@
...
+          ticker.last      = NumericHelper.to_d(ticker_json['last'])
+          ticker.volume    = NumericHelper.to_d(ticker_json['vol'])
+          ticker.timestamp = ticker_json['at'].to_i
+          ticker.payload   = output
+          ticker
+        end
+
+        private
+
+      def fetch_inner(endpoint)
```
tmlee: This has already been implemented in the parent class. Shouldnt need to worry about this
Also noticed that https://graviex.net/api/v2/tickers.json returns the full tickers.
I think that would be preferred over calling 1 ticker at a time. Save calls, less hit onto the exchange
@demuskov

tmlee:
@demuskov Thank you for your contribution. Just a couple of comments up there you might want to look into

vitorgamer58:
Gravies is compatible with an API that returns all tickets, it's even better if a new currency is added, that's how I tried to add Graviex to my pull request https://github.com/coingecko/cryptoexchange/pull/421

lib/cryptoexchange/exchanges/graviex/services/market.rb
demuskov: We neeed SSL_VERIFY turned off becouse of incomplete certificate issued by COMODO. We already in contact with them. But there is no issue at all - comm channel encrypted as should be.

lib/cryptoexchange/exchanges/graviex/graviex.yml
demuskov: yml file not needed - i'v rewrite code for direct pairs listing from API

demuskov:
So, i'v checked all of functionality, so:
1. All code is working as well
2. yml file is redundant and was initially composed by @vitorgamer58

vitorgamer58:
TOP

demuskov:
@tmlee when our code will be accepted/reviewed? Or may be there is some unclear moments or something? When we can be listed?

vitorgamer58:
a essa altura alguém já poderia ter aprovado o pull request

tmlee:
Sorry for the delay. I'll be accepting this pull request for now.
However, the `ticker` method should preferably use `https://graviex.net/api/v2/tickers/` instead of a single request for each pair; this will reduce load on the server and also reduces chance of hitting rate limit

PR #457, opened 2018-04-23
- Created market pair, orderbook, trades and ticker. Rspec has bene i…

cySheng:
The purpose of this pull request is to integrate coinut api.
Issue #383 
- [X] I have added Specs

1. I have changed `coinut_auth.yml` to be `cryptoexchange_api_keys.yml` instead. Definitely seems like a better option for the future if there is any similar APIs out there.

I have also added a pending method to the rspec test to ensure travis ci pass. Let me know if anything needs to be changed.

2. Through some bypass, I am also able to generate the cassettes for orderbook and trades. 
However, there is something wrong with the rspec test for both order book and trades. 

For, order_book the output should be something like this:
```
{"buy"=>
  [{"count"=>2, "price"=>"10748.18", "qty"=>"0.03700000"},
   {"count"=>1, "price"=>"10732.33", "qty"=>"0.00883000"},
   {"count"=>1, "price"=>"10694.94", "qty"=>"0.00261000"},
```

However, it emulates the previous request which is the ticker call instead, returning this: 

```
{"high24"=>"11876.26000000",
 "highest_buy"=>"11663.35000000",
 "inst_id"=>852380,
 "last"=>"11876.26000000",
 "low24"=>"10002.77000000",
 "lowest_sell"=>"11691.86000000",
 "nonce"=>29103,
 "open_interest"=>"0.00000000",
 "prev_trans_id"=>2421767187,
 "price_change_24"=>"0.17433023",
 "reply"=>"inst_tick",
 "status"=>["OK"],
 "timestamp"=>1524464445296641,
 "trans_id"=>2421767632,
 "volume"=>"3.13233885",
 "volume24"=>"0.06440000",
 "volume24_quote"=>"713.20039346",
 "volume_quote"=>"31239.97515938"}
```
Running it in binding.pry and calling the code manually `prepare_and_send_request` will allow the test to pass correctly, which is very weird. 
I suspect that its due to the nonce that is generated but I cannot confirm.

Will have a deeper look into it when time allows it.


lib/cryptoexchange/services/pairs.rb
```diff
@@ -38,19 +38,32 @@ def fetch_via_api(endpoint = self.class::PAIRS_URL)
...
       end
 
-      def fetch_via_api_using_post(endpoint = self.class::PAIRS_URL)
-        fetch_response = HTTP.timeout(:write => 2, :connect => 5, :read => 8).post(endpoint)
+      def fetch_via_api_using_post(endpoint = self.class::PAIRS_URL, headers = false, body = false)
+        fetch_response = 
+          if headers && body
+            HTTP.timeout(:write => 2, :connect => 5, :read => 8).headers(headers).post(endpoint, body: body)
+          else
+            HTTP.timeout(:write => 2, :connect => 5, :read => 8).post(endpoint)
```
dominicwong617: @cySheng the only difference between this method and `fetch_using_post` is to not have `:json => params`?

cySheng: I don't seem to understand what you're question is referring to. The coinut api requires certain params to be sent in the headers and body. If the headers and body is present then attach it to the post request.

Issue #466, opened 2018-04-25
Implement f1cx Ticker, Orderbook, Trades


TJ-CoinGecko:
website: https://f1cx.com/
https://f1cx.com/documents/api_v2

wenjieyek:
I will do this

PR #473, opened 2018-04-27
add f1cx exchange

wenjieyek:
- [x] I have added Specs


lib/cryptoexchange/exchanges/f1cx/services/market.rb
```diff
@@ -0,0 +1,38 @@
+module Cryptoexchange::Exchanges
+  module F1cx
+    module Services
+      class Market < Cryptoexchange::Services::Market
+        class << self
+          def supports_individual_ticker_query?
+            true
```
tmlee: You can actually do it in a single query with https://f1cx.com/api/v2/tickers
we can save api calls
@wenjieyek

lib/cryptoexchange/exchanges/f1cx/services/order_book.rb
```diff
@@ -0,0 +1,45 @@
...
+
+        def adapt(output, market_pair)
+          order_book = Cryptoexchange::Models::OrderBook.new
+
+          order_book.base      = market_pair.base
+          order_book.target    = market_pair.target
+          order_book.market    = F1cx::Market::NAME
+          order_book.asks      = adapt_orders(output['asks'])
+          order_book.bids      = adapt_orders(output['bids'])
+          order_book.timestamp = Time.now.to_i
```
tmlee: should we map this to `created_at` ? @wenjieyek

tmlee:
https://github.com/coingecko/cryptoexchange/issues/466

PR #479, opened 2018-04-27
handling gzip response

Iworb:
For further work it should be implemented if server returns only gzipped data.

### DIFF
--- a/lib/cryptoexchange/services/market.rb
+++ b/lib/cryptoexchange/services/market.rb
@@ -1,3 +1,5 @@
+require 'zlib'
+
 module Cryptoexchange
   module Services
     class Market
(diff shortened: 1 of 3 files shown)


Iworb:
Looks like there's a single site (Latoken) who sends header 'Content-Encoding' => 'gzip' withot actually gzipping content.

tmlee:
@Iworb Will you be able to include a spec for Latoken to tackle this scenario?

Iworb:
Well, it will be wrong if I include catching for `latoken` only. It depends on server response with wrong headers: content is not gzipped, but response header says it is. I don't know was there a temporary error or constant, could we re-run this spec?

Iworb:
`rspec spec/echanges/latoken` works fine. Could it be some temporary server-side error?
Is there a way to re-run Travis check?
image

Issue #489, opened 2018-05-04
Implement token.store Ticker, Orderbook, Trades


bobbyong:
https://token.store/API
https://github.com/tokenstore/apidocs

Issue #406, opened 2018-04-06
Implement The Token Store Ticker, Orderbook, Trades


TJ-CoinGecko:
website: https://token.store

estherleongym:
I'll take this

PR #522, opened 2018-05-21
The Token Store Integration

estherleongym:
- Added Pairs, Tickers, OrderBook, Trades and updated README for The Token Store
- Related to Issue #406 & #489 
- [ X ] I have added Specs


lib/cryptoexchange/exchanges/thetokenstore/services/market.rb
```diff
@@ -0,0 +1,49 @@
...
+
+        def adapt(market_pair, output)
+          ticker = Cryptoexchange::Models::Ticker.new
+          ticker.base = market_pair.base
+          ticker.target = market_pair.target
+          ticker.market = Thetokenstore::Market::NAME
+          ticker.last = NumericHelper.to_d(output[1]['last'])
+          ticker.ask = NumericHelper.to_d(output[1]['ask'])
+          ticker.bid = NumericHelper.to_d(output[1]['bid'])
+          ticker.volume = NumericHelper.to_d(output[1]['baseVolume'])
```
tmlee: @estherleongym mind double confirming the volume? I think we should take the other one. Our volume is in our base not theirs

estherleongym: @tmlee sorry! I overlooked this, it was suppose to point at quoteVolume instead since base and volume is swapped for this API. I've pushed the change

Issue #562, opened 2018-05-28
Implement JEX Ticker, Orderbook, and Trades


TJ-CoinGecko:
www.jex.com
https://www.jex.com/api/v2/pub/jexcap

danielchengml:
I'll take this

PR #572, opened 2018-05-30
Implements JEX Exchange: Ticker and MarketPairs, Update README, #562

danielchengml:
- Implemented Pairs, Tickers and updated README for JEX
- Related to Issue #562
- [x] I have added Specs


lib/cryptoexchange/exchanges/jex/services/market.rb
```diff
@@ -0,0 +1,57 @@
...
+        end
+
+        def ticker_url
+          "#{Cryptoexchange::Exchanges::Jex::Market::API_URL}"
+        end
+
+        def adapt_all(output)
+          market_pairs = []
+          output.map do |pair, ticker|
+            separator = /( CALL | PUT |ETF)/ =~ pair
```
tmlee: since these are blacklist keywords, maybe call it something other than separator?

lib/cryptoexchange/exchanges/jubi/services/market.rb
```diff
@@ -21,10 +21,10 @@ def adapt_all(output)
           default_target = 'CNY'
           output.map do |base, ticker|
             market_pair = Cryptoexchange::Models::MarketPair.new(
-                            base: base,
-                            target: default_target,
-                            market: Jubi::Market::NAME
-                          )
```
tmlee: no changes for jubi

lib/cryptoexchange/exchanges/jex/services/market.rb
danielchengml: Noted. I have resolved this issue by creating a method to check derivatives and remove it from the array. Please find the latest PR for your review.

danielchengml:
Hi @tmlee I have updated the code by doing the following: 

- adding test for derivatives (test: 'check derivatives removed') 
- cleaning up the naming conventions (removing "separator" and using an operator `if !derivative(pair)` to exclude derivatives from the return array.
- Writing a separate method to check for derivatives `def derivative(pair)`

Issue #492, opened 2018-05-10
Implement stellarport.io Ticker, Orderbook, Trades


TJ-CoinGecko:
website: portal.stellarport.io/exchange
https://stellarportio.zendesk.com/hc/en-us/articles/360001050894-Top-Markets

danielchengml:
I'll take this

PR #580, opened 2018-06-04
Implement Stellarport Ticker & Market Pairs

danielchengml:
- What is the purpose of this Pull Request? As Above
- What is the related issue for this Pull Request? #492 
- [x] I have added Specs

Note: 
- 
- All the base currency is in "XLM" while the target varies.
- The volume returned from the API consists of `soldVolume` and `boughtVolume`. I combined them together to show a single `volume` entry.
- API doesn't return `bid` and `ask` price.
- As of late, there are 48 trading pairs listed. API claims to only show the top 100 pairs. Should the pair increase to >100, this API might have to be revised.


istrau2:
Hi @danielchengml ,

I am from Stellarport. I was told to comment here regarding potential conflicts with Stellarport integration.

I am not sure which conflicts I am supposed to be commenting on. However, I will say that we have a new updated ticker endpoint that this PR does not look to be using. Please see these links:
https://stellarportio.zendesk.com/hc/en-us/articles/360001050894-Ticker-Data
https://stellar.api.stellarport.io/Ticker

Issue #589, opened 2018-06-04
Implement Simex Ticker, Orderbook, and Trades


bobbyong:
https://simex.global/en
https://simex.global/en/docs/introduction

danielchengml:
I'll take this

PR #609, opened 2018-06-06
Implement Simex Ticker

danielchengml:
- What is the purpose of this Pull Request? As above
- What is the related issue for this Pull Request? #589 
- [x] I have added Specs


tmlee:
@danielchengml should probably change the User git configuration to the same email as github. As you can see here Github cannot refer the commit to your github account

Issue #653, opened 2018-06-27
Implement OmniTrade Ticker, Orderbook, and Trades


TJ-CoinGecko:
https://omnitrade.io/
https://omnitrade.io/documents/api_v2

PR #656, opened 2018-06-27
OmniTrade integration

laka3000:
- Integrate OmniTrade exchange. Pairs, Tickers, OrderBook, Trades and README updated. 
- Related to issue #653 
- [x] I have added Specs


tmlee:
@laka3000 Thank you for your contribution!

Issue #632, opened 2018-06-18
Implement MySpeedTrade Ticker, Orderbook, and Trades


TJ-CoinGecko:
https://myspeedtrade.com/
https://myspeedtrade.com/documents/api_v2?lang=en

globalpaynet:
What you need to give to speed up the listing process

MySpeedTrade:
the platform is same as Graviex #453

estherleongym:
I'll take this

tmlee:
Unresolved

Getting an issue of `Cryptoexchange::ResultParseError: {:response=>"General no method error thrown"}` while testing

PR #661, opened 2018-06-28
MySpeedTrade Integration

estherleongym:
- Implemented Pairs, Tickers, OrderBook, and updated README for MySpeedTrade
- Related to issue #632 
- [ X ] I have added Specs


estherleongym:
Did not implement Trades because 'side' currently returns 'null'.

Issue #570, opened 2018-05-30
Implement Liqnet Ticker, Orderbook, and Trades 


TJ-CoinGecko:
https://liqnet.com/exchange/
https://liqnet.com/public-api/special/

estherleongym:
I'll take this

PR #668, opened 2018-06-30
Liqnet Integration

estherleongym:
- Implemented Pairs, Tickers, OrderBook, Trades, and updated README for Liqnet
- Related to issue #570 
- [ X ] I have added Specs

### DIFF
--- a/lib/cryptoexchange/exchanges/liqnet/services/id_fetcher.rb
+++ b/lib/cryptoexchange/exchanges/liqnet/services/id_fetcher.rb
@@ -1,0 +1,20 @@
+module Cryptoexchange::Exchanges
+  module Liqnet
+    module Services
+      class IdFetcher
+        PAIR_SYMBOLS_URL = "#{Cryptoexchange::Exchanges::Liqnet::Market::API_URL}/markets"
+        PAIR_SYMBOLS = []
+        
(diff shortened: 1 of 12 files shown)


lib/cryptoexchange/exchanges/liqnet/services/id_fetcher.rb
```diff
@@ -0,0 +1,20 @@
...
+  module Liqnet
+    module Services
+      class IdFetcher
+        PAIR_SYMBOLS_URL = "#{Cryptoexchange::Exchanges::Liqnet::Market::API_URL}/markets"
+        PAIR_SYMBOLS = []
+        
+        def self.get_id(market_pair)
+          if PAIR_SYMBOLS.empty?
+           pairs_response = HTTP.get(PAIR_SYMBOLS_URL)
+           PAIR_SYMBOLS << pairs_response.parse(:json)
```
wongy91: this will cause issue if the constant is not being cleared (or assign a new client), perhaps dont 'cache' this for now since this is only required by OrderBooks and put a TODO comment.

estherleongym: ok got it, will fix it to not store for now + add the comment

PR #689, opened 2018-07-11
added coindcx exchange

wenjieyek:
- [ ] I have added Specs


lib/cryptoexchange/exchanges/coindcx/services/market.rb
```diff
@@ -0,0 +1,53 @@
...
+          adapt_all(output)
+        end
+
+        def ticker_url
+          "#{Cryptoexchange::Exchanges::Coindcx::Market::API_URL}/exchange/ticker"
+        end
+
+        def adapt_all(output)
+          output.map do |ticker|
+            pair = ticker['market'].insert(-4, '/')
```
tmlee: This approach is a bit dangerous, try something like https://github.com/coingecko/cryptoexchange/blob/master/lib/cryptoexchange/exchanges/abcc/services/pairs.rb

Issue #638, opened 2018-06-20
Implement CredoEx Ticker, Orderbook, and Trades


TJ-CoinGecko:
https://credoex.com
https://credoex.com/tickers

estherleongym:
I'll take this

PR #695, opened 2018-07-16
CredoEx Integration

estherleongym:
- Implemented Pairs and Tickers, updated README for CredoEx
- Related to issue #638 
- [ X ] I have added Specs


MartyBauer:
Hey checking when this will be deployed

Issue #369, opened 2018-03-14
Implement Coinhub Ticker, Orderbook, Trades 


bobbyong:
https://www.coinhub.io/gateway/coinhub/public/returnTicker

estherleongym:
API endpoints returning error message; reached out to Coinhub to verify.

**Sample:** https://www.coinhub.io/api#!/exchange.json/chart

**Request sent**: `https://www.coinhub.io/gateway/public/chart?market="ETH-USD"&openIntervals	=true`

**Response returned**: `{"success":false,"missing_authenticators":[],"infos":[],"warnings":[],"errors":[{"JSON_DECODE_FAILED":["No error"]}],"results":{}}`

Swingcloud:
let me do this

estherleongym:
I'll take this, discussed with @Swingcloud already

TJ-CoinGecko:
Provided by coinhub: https://www.coinhub.io/api#/cnh

PR #734, opened 2018-07-26
Coinhub Integration

estherleongym:
- Implemented Pairs, Tickers, and Trades for Coinhub
- Related to issue #369 
- [ X ] I have added Specs


wongy91:
@estherleongym can you help rebase against master? there are conflicts

PR #834, opened 2018-08-12
Tidied Exchange table in README.md

alphafoobar:
- Exchanges supported table was incorrectly formatted for online rendering.
- Small change, no issue created.
- [N/A ] I have added Specs
- [N/A] (If implementing Market Ticker) I have verified that the `volume` refers to BASE
- [N/A] (If implementing Market Ticker) I have verified that the `base` and `target` is assigned correctly


tmlee:
Woops didnt see this PR, noticed the table scrambled and fixed it separately. Thanks for spotting it though

Issue #848, opened 2018-08-16
Tradeogre url base/target is inverted 


TJ-CoinGecko:
Eg: https://www.coingecko.com/en/coins/electronero
Existing wrong implementation: https://tradeogre.com/exchange/ETNX-LTC
Correct implementation should be: https://tradeogre.com/exchange/LTC-ETNX

PR #850, opened 2018-08-17
swap trade-ogre url base and target

Swingcloud:
- What is the related issue for this Pull Request? #848


ernsheong:
@Swingcloud in the future would be nice to put Fixes #848 in the PR description so that the issue auto-closes :)

Swingcloud:
hello @ernsheong seems I have put issues id on PR description, but seems not work, maybe I need to put on commit message?

PR #878, opened 2018-08-27
AXNET Exchange Integration

superdev810:
- Integrated Ticker, Orderbook, Trades and Update Readme
- No related any issue
- [ X ] I have added Specs
- [ X ] (If implementing Market Ticker) I have verified that the `base` and `quote` is assigned correctly


tmlee:
Thanks for contribution, but there seem to be an issue with this implementation https://github.com/coingecko/cryptoexchange/issues/879#issuecomment-416803935

PR #909, opened 2018-09-04
Bitbox updated API endpoints authenticated

tmlee:
Support Bitbox authenticated API

This CI will fail for now


tmlee:
We are going to skip bitbox test for now

Issue #873, opened 2018-08-27
[Allcoin] updated API, existing API not usable.


wongy91:
existing API endpoints not returning data data anymore, apparently they have a new endpoint. https://www.allcoin.com/api_market/market

estherleongym:
I'll take this

PR #943, opened 2018-09-12
Update Allcoin API & Add Pairs and Trades

estherleongym:
- Updated Allcoin API endpoint and README & added Pairs and Trades
- Fixes: Related to issue #873 
- [ X ] I have added Specs


wongy91:
tested on local, LGTM.

Issue #935, opened 2018-09-12
Implement oax


TJ-CoinGecko:
www.oax.com
http://www.oax.com/#/help/help_detail/20/4

estherleongym:
I'll take this

PR #977, opened 2018-09-27
OAX Integration

estherleongym:
- Added Pairs, Tickers, OrderBook, Trades, and updated README for OAX
- Closes: #935 
- [ X ] I have added Specs


wongy91:
compared the volume, looks right. (volume in API is base volume, volume in exchange page is showing ETH volume)

PR #1004, opened 2018-10-04
update readme, fix UEX timestamp.

wongy91:
update missing readme, fix timestamp

- [x] I have added Specs


tmlee:
There should be a timestamp spec to see if its within standard @wongy91

PR #1008, opened 2018-10-06
WIP - WebSocket PoC with Bitfinex

dominicwong617:
@tmlee let me know what you think about this


tmlee:
Does bitfinex support last trade as well? or is this reflected in the ticker itself when a latest trade kicks in

README.md
```diff
@@ -310,6 +310,20 @@ Or install it yourself as:
...
 ```
 
+### Query the Ticker stream API
+
+```
+  pair = client.pairs('bitfinex').first
+  thread = client.ticker_stream(
+    market_pair: pair,
+    onopen: Proc.new { puts 'Opened!' },
+    onmessage: Proc.new { |ticker| puts do_something_with_ticker(ticker) },
```
tmlee: I would put a working example here to print the ticker value rather than `do_something_with_ticker`

In fact when `do_something_with_ticker` was used, there was no indication of any error thrown. No mention of undefined method etc.

lib/cryptoexchange/client.rb
```diff
@@ -80,5 +80,69 @@ def trades(market_pair)
...
+        EM.run do
+          ws = WebSocket::EventMachine::Client.connect(uri: stream.url)
+
+          ws.onopen do
+            onopen.call if onopen
+            ws.send(stream.subscribe_event(market_pair))
+          end
+
+          ws.onmessage do |message, _|
+            message = JSON.parse(message)
```
tmlee: We might want to print the message here for debugging purposes and possible future error handling

lib/cryptoexchange/client.rb
```diff
@@ -80,5 +80,69 @@ def trades(market_pair)
...
+
```
tmlee: There should also be a way to tell why a WS socket closed in development. Otherwise developers may be lost just because their code had a typo in the block, ws closed

PR #1055, opened 2018-10-11
add incorex api

osivinyuk:
- We want to add IncoreX exchange to your lib?
-  I have added Specs
- I have verified that the `volume` refers to BASE
- I have verified that the `base` and `target` is assigned correctly


estherleongym:
Hi @osivinyuk, could you rebase master and push the updated branch again. This is to fix the failing CIs.

Issue #1024, opened 2018-10-09
Implement Ccore Exchange


TJ-CoinGecko:
https://ccore.io/exchange?Id=15
https://websitewebapi.ccore.io/integration/dailyhistory?partnerId=2

PR #1063, opened 2018-10-13
Ccore.io Integration

estherleongym:
- Added Pairs, Tickers, and updated README for Ccore
- Closes: #1024 
- [X] I have added Specs


wongy91:
has 0 volume. unable to verify ticker volume.

Issue #1040, opened 2018-10-09
Implement ProBitEX Exchange


TJ-CoinGecko:
https://www.probitex.com
https://www.probitex.com/api

Swingcloud:
I'll do this

probitex:
Updated NEW API link is: 
https://www.probitex.com/apis
 Only sub-folder changed from api to apis

probitex:
Public API V.5.0(Updated: 08.05.2019)

 
GET Public Single Market Pair Stats
https://api.probitex.com/apis/get_market?name=ltc_btc
 
GET Public Full Market Stats
https://api.probitex.com/apis/get_market_all
 
GET Public Specified market depth
https://api.probitex.com/apis/get_market_depth?name=ltc_btc
 
GET Public Specified market Orderbook
https://api.probitex.com/apis/get_market_logs?name=ltc_btc

PR #1112, opened 2018-10-27
✨ Implement Probitex Ticker and orderbook 

Swingcloud:
- What is the related issue for this Pull Request (if this PR fixes issue, prepend with "Fixes" or "Closes")? #1040
- [x] I have added Specs

```
```

ticker on `LTC_BTC`
url: https://www.probitex.com/trade/index/market/ltc_btc
```ruby
{
 "ltc_btc": {
  "new_price": "0.00813000",
  "buy_price": "0.00091392",
  "sell_price": "0.09110804",
  "min_price": "0.00790000",
  "max_price": "0.00813000",
  "24h_num": "765.51000000",
  "24h_mum": "6.13634790",
  "24h_float": "1.37"
 }
}
```
took `24h_num` as base volume
and seems 24h_float is `change` but when it's in negative value, it will return false, so didn't implement that value

in `trades` endpoint
https://www.probitex.com/apis/get_market_logs?name=ltc_btc
that timestamp only return time, so didn't implement it in this PR


lib/cryptoexchange/exchanges/probitex/services/market.rb
```diff
@@ -0,0 +1,44 @@
...
+          ticker.target    = target
+          ticker.market    = Probitex::Market::NAME
+          ticker.bid       = NumericHelper.to_d(output['buy_price'])
+          ticker.ask       = NumericHelper.to_d(output['sell_price'])
+          ticker.high      = NumericHelper.to_d(output['max_price'])
+          ticker.low       = NumericHelper.to_d(output['min_price'])
+          ticker.volume    = NumericHelper.to_d(output['24h_num'])
+          ticker.timestamp = nil
+          ticker.payload   = [pair, output]
+          ticker
```
wongy91: there is no `last`. I think can use the json attribute `new_price`

Swingcloud: will do

Swingcloud: @wongy91 I've updated! you might take a look

PR #1175, opened 2018-11-22
Alien Cloud Integration

estherleongym:
- Add Pairs, Tickers, OrderBook, Trades, Trade URL and updated README for AlienCloud
- Closes: #1169 
- [X] I have added Specs

```
```


affka:
Hello, @estherleongym ! We need completion this integration, can you continue to do it? We can pay you work. How to contact you?

Issue #1170, opened 2018-11-21
Implement CryptoExchange.ws


TJ-CoinGecko:
https://www.crypto-exchange.ws/
https://cryptoexchange.ws/api
https://cryptoexchange.ws/en/gateway/public/returnTicker
https://www.crypto-exchange.ws/Coin-Info/

estherleongym:
I'll take this

CryptoExchangews:
Dear Team, I'm not expert on github. How is the status of cryptoexchange.ws integration? All the documents are ok? Could you tell me how many days needed to get the full listing on coingecko?
thank you for the support.

PR #1179, opened 2018-11-25
Cryptoexchange.ws Integration

estherleongym:
- Add Pairs, Tickers, OrderBook, Trades, Trade URL and updated README for Cryptoexchange.ws
- Closes: #1170 
- [X] I have added Specs

```
```


lib/cryptoexchange/exchanges/cryptoexchange_ws/market.rb
```diff
@@ -0,0 +1,12 @@
+module Cryptoexchange::Exchanges
+  module CryptoexchangeWs
+    class Market < Cryptoexchange::Models::Market
+      NAME = 'cryptoexchange_ws'
```
CryptoExchangews: name = 'CryptoExchange.ws'

Issue #1178, opened 2018-11-24
Implement Swiftex


bobbyong:
https://swiftex.co/
https://swiftex.co/documents/api_v2?lang=en

estherleongym:
I'll take this

PR #1190, opened 2018-11-29
Swiftex Integration

estherleongym:
- Add Pairs, Tickers, OrderBook, Trade URL and updated README for Swiftex
- Closes: #1178 
- [X] I have added Specs

```
```


joenilan:
just wondering if this is getting integrated eventually? they have an updated api, would be great to see them on. 

https://documenter.getpostman.com/view/6576473/RztoKnmP

Issue #1211, opened 2018-12-07
Implement Poloniex Websocket API for Trades, Orderbook and Ticker


srisankethu:
https://poloniex.com/support/api/#websockets

PR #1212, opened 2018-12-07
WIP: Implement Poloniex websocket API

srisankethu:
Solves https://github.com/coingecko/cryptoexchange/issues/1211
- [ ] I have added Specs
- [ ] I have verified at least **ONE** ticker volume matches volume shown on the trading page (use script below)

```
```


srisankethu:
Will update it in 5-7 days :+1:

tmlee:
@srisankethu 

- I tested this code but it seems like the values are not getting mapped correctly, everything seems blank other than `base` and `target`. I supposed it is pointing to the wrong element in the array.
eg. the ticker message is in the following format
```
[1002, nil, [50, "0.00830500", "0.00831376", "0.00830133", "0.02792413", "111.12106534", "13454.60628988", 0, "0.00846000", "0.00802279"]]
```

- Since this websocket returns all pairs in a single connection, it would be ideal to have it done in a way that we do not need to pass in a specific pair.
In any case a mapping needs to be done with (Refer to the bottom currency pair ID at https://poloniex.com/support/api/#websockets_channels_ticker) whichever approach we choose to take. Don't seem to see any API that allows us to pull this list in real time

- Would also suggest documenting the websocket output response in the comment (see Bitfinex implementation) as a reference for others.

srisankethu:
@tmlee Thanks for the comment. I will look into it ASAP.

srisankethu:
@tmlee I am facing problem with the ticker format. Unlike Bitfinex, this is a different format. How do I test my implementation?

PR #1221, opened 2018-12-19
Alterdice

Alterdice:
Change api url


wongy91:
why is the URL change needed? any breaking (or non-breaking) changes from alterdice side that we should know?

wongy91:
@Alterdice thanks for the PR however there's a conflict, can you help resolve that. thanks

Issue #762, opened 2018-07-30
Implement LocalTrade Ticker, Orderbook, Trades


TJ-CoinGecko:
https://localtrade.cc/

wongy91:
api: https://localtrade.cc/main/ticker/

estherleongym:
I’ll take this

PR #1268, opened 2019-01-08
LocalTrade Integration

estherleongym:
- Add Pairs, Tickers, Trade URL and updated README for Localtrade.cc
- Closes: #762 
- [X] I have added Specs
**- [ ] I have verified at least **ONE** ticker volume matches volume shown on the trading page (use script below)** _Unable to verify volume against website - not able to locate on trading page_


wongy91:
@estherleongym spec is failing for all 3 builds, kindly check

PR #1361, opened 2019-02-12
Elvis/saturn network intergration

ElvisLYC:
- [x] I have added Specs

```
```


lib/cryptoexchange/exchanges/saturn_network/services/pairs.rb
```diff
@@ -0,0 +1,26 @@
...
+
+        def fetch
+          output = super
+          adapt(output)
+        end
+
+        def adapt(output)
+          output.map do |pair|
+            base = pair[0].split('_').first
+            target = pair[1]["symbol"]
```
tmlee: In our convention target is usually the dominant currency like USD, BTC, in this case ETH/ETC
base is the currency that varies

lib/cryptoexchange/exchanges/saturn_network/services/market.rb
```diff
@@ -0,0 +1,50 @@
...
+        def adapt(output, market_pair)
+          ticker = Cryptoexchange::Models::Ticker.new
+
+          ticker.base      = market_pair.base
+          ticker.target    = market_pair.target
+          ticker.market    = SaturnNetwork::Market::NAME
+          ticker.last      = NumericHelper.to_d(output[1]["last"])
+          ticker.high      = NumericHelper.to_d(output[1]["highestBid"])
+          ticker.low       = NumericHelper.to_d(output[1]["lowestAsk"])
+          ticker.volume    = NumericHelper.to_d(output[1]["baseVolume"])
```
tmlee:
```

name | "Saturn Classic DAO Token"
-- | --
symbol | "SATURN"
quoteVolume | "30000"
baseVolume | "1.45"
highestBid | "0.00003"
lowestAsk | "0.00005"
last | "0.00005"
percentChange | "0.06230889172387269"
```

From the example response. We want the volume to be in base in our convention.
If you fix the base/target by reversing it (https://github.com/coingecko/cryptoexchange/pull/1361/files#r255827784), then you will want to take quoteVolume

tmlee: You need to reverse it because otherwise it would be ETH/SATURN
And by assigning `last` it would be deem as 0.0003 SATURN per ETH

Issue #423, opened 2018-04-13
Implement BitShares Asset Exchange Protocol Ticker, Orderbook, Trades


bobbyong:

Swingcloud:
this maybe the api doc 
http://docs.bitshares.org/api/

tmlee:
Lets use this API by http://cryptofresh.com/api/docs

Track BTS target only for now
https://cryptofresh.com/api/asset/markets?asset=BTS

Swingcloud:
👌

wenjieyek:
https://github.com/coingecko/cryptoexchange/pull/1366

PR #1366, opened 2019-02-15
implement bitshares assets

wenjieyek:
- Closes: https://github.com/coingecko/cryptoexchange/issues/423
- [x] I have added Specs

```
```


lib/cryptoexchange/exchanges/bitshares_assets/services/market.rb
```diff
@@ -0,0 +1,35 @@
+module Cryptoexchange::Exchanges
+  module BitsharesAssets
+    module Services
+      class Market < Cryptoexchange::Services::Market
+        class << self
+          def supports_individual_ticker_query?
+            true
```
tmlee: Because https://cryptofresh.com/api/asset/markets?asset=BTS returns a list of tickers, this would be a `false` example right?

lib/cryptoexchange/exchanges/bitshares_assets/services/market.rb
```diff
@@ -0,0 +1,35 @@
...
+          "#{Cryptoexchange::Exchanges::BitsharesAssets::Market::API_URL}/asset/markets?asset=BTS"
+        end
+
+        def adapt(output, market_pair)
+          ticker           = Cryptoexchange::Models::Ticker.new
+          ticker.base      = market_pair.base
+          ticker.target    = market_pair.target
+          ticker.market    = BitsharesAssets::Market::NAME
+
+          ticker.last      = NumericHelper.to_d(output['OPEN.BTC']['price'])
```
tmlee: Doing this would mean that it is only taking a value for `OPEN.BTC`, this shouldnt be hardcoded since there are many other currencies in the response

lib/cryptoexchange/exchanges/bitshares_assets/services/market.rb
```diff
@@ -0,0 +1,35 @@
...
+          end
+        end
+
+        def fetch(market_pair)
+          output = super(ticker_url)
+          adapt(output, market_pair)
+        end
+
+        def ticker_url
+          "#{Cryptoexchange::Exchanges::BitsharesAssets::Market::API_URL}/asset/markets?asset=BTS"
```
wongy91: we only want `BTS` ?

tmlee: BTS only as starter, otherwise too many pairs

PR #1429, opened 2019-03-15
Elvis/bitmesh exchange intergration

ElvisLYC:
- What is the related issue for this Pull Request (if this PR fixes issue, prepend with "Fixes" or "Closes")?closes#1426
- [x] I have added Specs

```
```


lib/cryptoexchange/exchanges/bitmesh/services/pairs.rb
```diff
@@ -0,0 +1,32 @@
...
+        def fetch
+          output = super
+          adapt(output)
+        end
+
+        def adapt(output)
+          pairs = output
+          market_pairs = []
+          pairs["data"].each do |pair|
+            unless pair[1]['price'].nil?
```
ElvisLYC: some ticker does not return `last`, so to exclude these tickers. example as below:-

```
        "eth_bch": {
            "name": "eth_bch",
            "change": 0
        },
```

lib/cryptoexchange/exchanges/bitmesh/services/market.rb
```diff
@@ -0,0 +1,49 @@
...
+          output = super(ticker_url)
+          adapt_all(output)
+        end
+
+        def ticker_url
+          "#{Cryptoexchange::Exchanges::Bitmesh::Market::API_URL}"
+        end
+
+        def adapt_all(output)
+          output["data"].map do |pair|
```
tmlee: iterate key,value so you dont have to pass in `[0]` below

tmlee: will merge and change it

PR #1584, opened 2019-06-09
FIX: corrects slicing of CoinMetro's currency pairs

BeardPower:
Fixes an issue with the slicing of currency pairs.


wongy91:
resolved in https://github.com/coingecko/cryptoexchange/pull/1585

PR #1721, opened 2019-08-26
fix bithumb volume (api updated)

wongy91:
bithumb updated their api apparently, json key for `volume` has changed


streetcrypto7:
Hello can you merge my PR? Thanks.

PR #1742, opened 2019-09-04
add_support_for_contract_stat

ElvisLYC:
I added `contract_stat` for two  futures exchanges (bybit and okex_swap) to test out two possible scenario. Please refer to my comments below.


lib/cryptoexchange/exchanges/bybit/services/contract_stat.rb
```diff
@@ -0,0 +1,51 @@
+module Cryptoexchange::Exchanges
```
ElvisLYC: this is a scenario where `open_interest` and `index` data came from the same `url`

lib/cryptoexchange/exchanges/okex_swap/services/contract_stat.rb
```diff
@@ -0,0 +1,41 @@
+module Cryptoexchange::Exchanges
```
ElvisLYC: this is a scenario where `open_interest` and `index` data  came from different url

spec/exchanges/bybit/integration/market_spec.rb
```diff
@@ -52,5 +52,18 @@
     expect(order_book.bids.count).to be > 0
     expect(order_book.timestamp).to be_nil
     expect(order_book.payload).to_not be nil
-  end  
+  end
+
+  it 'fetch contract stat' do
```
ElvisLYC: added spec for `contract_stat`

lib/cryptoexchange/exchanges/okex_swap/services/contract_stat.rb
```diff
@@ -0,0 +1,41 @@
...
+        def fetch(market_pair)
+          unless market_pair.contract_interval == "perpetual"
+            open_interest = super(open_interest_url(market_pair))
+            index = super(index_url(market_pair))
+            adapt(open_interest, index, market_pair)
+          end
+        end
+
+        def open_interest_url(market_pair)
+          "https://www.okex.com/api/futures/v3/instruments/#{market_pair.inst_id}/open_interest"
```
tmlee: If the URL domain/prefix are shared from the `market.rb`,you can use that constant

lib/cryptoexchange/exchanges/bybit/services/market.rb
```diff
@@ -27,7 +27,7 @@ def adapt_all(output)
             market_pair = Cryptoexchange::Models::MarketPair.new(
               base: base,
               target: target,
-              market: Btcnext::Market::NAME
+              market: Bybit::Market::NAME
```
tmlee: is this an unrelated mistake ?

ElvisLYC: yes noticed this mistake from past implementation

lib/cryptoexchange/exchanges/okex_swap/services/contract_stat.rb
ElvisLYC: they are not the same.. anyway, i have assign them to a `CONTRACT_STAT_URL` as it is repetitive

lib/cryptoexchange/models/contract_stat.rb
```diff
@@ -0,0 +1,26 @@
...
+  module Models
+    class ContractStat
+      attr_accessor :base, :target, :market, :open_interest,
+                    :index, :timestamp, :payload
+
+      def initialize(args = {})
+        @base = args[:base]
+        @target = args[:target]
+        @market = args[:market]
+        @open_interest = args[:open_interest] || []
```
wongy91: why not just return nil?

lib/cryptoexchange/models/contract_stat.rb
```diff
@@ -0,0 +1,26 @@
...
+        @index = args[:index] || []
```
wongy91: same here

lib/cryptoexchange/models/contract_stat.rb
ElvisLYC: noted and changed

PR #1792, opened 2019-09-26
Contract interval fix

ElvisLYC:


lib/cryptoexchange/exchanges/coinflex/services/market.rb
```diff
@@ -42,6 +43,7 @@ def adapt(output, market_pair)
           ticker           = Cryptoexchange::Models::Ticker.new
           ticker.base      = market_pair.base
           ticker.target    = market_pair.target
+          ticker.contract_interval    = market_pair.contract_interval
```
tmlee: @ElvisLYC 
This is coinflex (spot) right? shouldnt need contract_interval
I just checked again the coinflex futures already have contract_interval

PR #1802, opened 2019-10-01
Add contract_stat info for more exchanges

ElvisLYC:


lib/cryptoexchange/exchanges/binance_futures/services/contract_stat.rb
```diff
@@ -9,26 +9,30 @@ def supports_individual_ticker_query?
...
           contract_stat.target    = market_pair.target
           contract_stat.market    = BinanceFutures::Market::NAME
-          contract_stat.index     = index['markPrice'].to_f
-          contract_stat.payload   = { "index" => index }
+          contract_stat.index     = contract_info['markPrice'].to_f
+          contract_stat.funding_rate_percentage     = contract_info['lastFundingRate'].to_f * 100
+          contract_stat.next_funding_rate_timestamp     = contract_info['nextFundingTime']/1000
+          contract_stat.funding_rate_percentage_predicted = nil
+          contract_stat.contract_type = "perpetual"
+          contract_stat.payload   = { "contract_info" => contract_info }
```
wongy91: curious why we want to wrap it in "contract_info"

lib/cryptoexchange/exchanges/kraken_future/services/contract_stat.rb
```diff
@@ -9,26 +9,72 @@ def supports_individual_ticker_query?
...
+        def process_contract_info(contract_info, market_pair)
+          arr = {}
+          if market_pair.contract_interval == "perpetual"
+            contract = contract_info['tickers'].find { |i| i['symbol'] == market_pair.inst_id.downcase }
+            arr["funding_rate_percentage"] = contract["fundingRate"].to_f * 100
+            arr["funding_rate_percentage_predicted"] = contract["fundingRatePrediction"].to_f * 100
+            arr["next_funding_timestamp"] = nil
+          else
+            contract = contract_info['instruments'].find { |i| i["symbol"] == market_pair.inst_id.downcase }
+            arr["expire_timestamp"] = DateTime.parse(contract["lastTradingTime"]).to_time.to_i
```
wongy91: curious do we still need `to_time` ?

wongy91:
@tmlee please check. looks fine to me, just a question, do we want to have validations on contract_stat, to validate a derivative ticker, and the type of derivative.

lib/cryptoexchange/exchanges/kraken_future/services/contract_stat.rb
ElvisLYC: yea, you need parse the `to_time` first before you can convert the time into timestamp or you will get `NoMethodError`

lib/cryptoexchange/exchanges/binance_futures/services/contract_stat.rb
ElvisLYC: actually we don't need to wrap it in `contract_info` in this case. The reason we wrap it in `open_interest`, `index` or `contract_info` in other cases is because the payload might came from different api endpoint it is easier to identify them this way. Will amend this.

tmlee:
@wongy91 what validations are you referring to?
Not too sure yet until we see it in action i think. We should find ways to keep track the attributes, will amend the table to be more granular

PR #1868, opened 2019-11-11
Fix ruby syntax warning - string literal in condition

teohm:
This PR removes the unusual appearance of "string literal" in `elsif` condition.

It causes an Ruby syntax warning:

```
cryptoexchange/exchanges/okex_swap/services/contract_stat.rb:31: warning: string literal in condition
```

### DIFF
--- a/lib/cryptoexchange/exchanges/okex_swap/services/contract_stat.rb
+++ b/lib/cryptoexchange/exchanges/okex_swap/services/contract_stat.rb
@@ -24,11 +24,12 @@ module Cryptoexchange::Exchanges
         end
 
         def get_contract_info(market_pair)
-          contract_info = if market_pair.contract_interval == "perpetual"
-                          HTTP.get("#{SWAP_URL}/#{market_pair.inst_id}/funding_time").parse(:json)
-                        elsif "futures"
-                          HTTP.get(FUTURES_URL).parse(:json)
-                        end
+          contract_info = \
+            if market_pair.contract_interval == "perpetual"
+              HTTP.get("#{SWAP_URL}/#{market_pair.inst_id}/funding_time").parse(:json)
+            else
+              HTTP.get(FUTURES_URL).parse(:json)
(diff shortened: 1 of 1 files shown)


lib/cryptoexchange/exchanges/okex_swap/services/contract_stat.rb
```diff
@@ -24,11 +24,12 @@ def fetch(market_pair)
         end
 
         def get_contract_info(market_pair)
-          contract_info = if market_pair.contract_interval == "perpetual"
-                          HTTP.get("#{SWAP_URL}/#{market_pair.inst_id}/funding_time").parse(:json)
-                        elsif "futures"
-                          HTTP.get(FUTURES_URL).parse(:json)
-                        end
```
teohm: It's unusual to do this in Ruby - `elsif "futures"`

tmlee: I think that is a mistake/bug left there, but suggesting at least 
`elsif market_pair.contract_interval == "futures"` 
just in case new products get added so we dont accidentally mistreat the result

teohm: @tmlee I wanted to do that at first. 

Then I saw `if contract_interval == "perpetual" .. else .. end` appears multiple times in the same file (see [L38](https://github.com/coingecko/cryptoexchange/blob/2e7394bd4eee1d8c113d8a87d3e8cda13e3285ba/lib/cryptoexchange/exchanges/okex_swap/services/contract_stat.rb#L38) and [L15](https://github.com/coingecko/cryptoexchange/blob/2e7394bd4eee1d8c113d8a87d3e8cda13e3285ba/lib/cryptoexchange/exchanges/okex_swap/services/contract_stat.rb#L15)). To make them consistent, I can update all together.

What do you think?

PR #1878, opened 2019-11-15
WIP: btse_futures index

ElvisLYC:


spec/exchanges/btse_futures/integration/market_spec.rb
```diff
@@ -80,11 +80,11 @@
       expect(contract_stat.base).to eq 'BTC'
       expect(contract_stat.target).to eq 'USD'
       expect(contract_stat.market).to eq 'btse_futures'
-      expect(contract_stat.index).to be nil
+      expect(contract_stat.index).to be_a Numeric
       expect(contract_stat.open_interest).to be_a Numeric
       expect(contract_stat.timestamp).to be nil
-      expect(contract_stat.index_identifier).to eq 'BtseFutures-BTC'
```
ElvisLYC: @tmlee how do you get this format of naming for `index_name` and `index_identifier`

lib/cryptoexchange/exchanges/btse_futures/services/contract_stat.rb
```diff
@@ -10,23 +10,31 @@ def supports_individual_ticker_query?
...
           output = output.find { |i| i["symbol"] == market_pair.inst_id }
+          index = index["data"].find { |i| i["id"] == market_pair.inst_id }
+
           contract_stat = Cryptoexchange::Models::ContractStat.new
 
           contract_stat.base      = market_pair.base
           contract_stat.target    = market_pair.target
           contract_stat.market    = BtseFutures::Market::NAME
           contract_stat.payload   = output
-          contract_stat.index_identifier = "BtseFutures-#{market_pair.base}"
```
ElvisLYC: shouldn't we pull directly from the index api? or is there a specific way to name these new attributes?

spec/exchanges/btse_futures/integration/market_spec.rb
tmlee: That is a placeholder for exchange that do not support index prior.
Since now this one has it, we should use its `id`

PR #2097, opened 2020-05-18
Update coss api

teekenl:
- [ ] I have added Specs
- [ ] I have verified at least **ONE** ticker volume matches volume shown on the trading page (use script below)

```
```


lib/cryptoexchange/exchanges/coss/services/market.rb
```diff
@@ -17,10 +17,9 @@ def ticker_url
           "#{Cryptoexchange::Exchanges::Coss::Market::API_URL}/ticker"
         end
 
-        def adapt_all(output)
-          result = output['result']
-          result.map do |value|
-            base, target = value['MarketName'].split('-')
+        def adapt_all(result)
+          result.map do |(market_name, value)|
```
wongy91, line 21: this is not incorrect, but the codebase for block, there is no bracket for the params.

lib/cryptoexchange/exchanges/coss/market.rb
```diff
@@ -2,10 +2,10 @@ module Cryptoexchange::Exchanges
   module Coss
     class Market < Cryptoexchange::Models::Market
       NAME = 'coss'
-      API_URL = 'https://exchange.coss.io/api'
+      API_URL = 'https://market.coss.io/api'
 
       def self.trade_page_url(args={})
-        "https://coss.io/c/trade?s=#{args[:base].upcase}_#{args[:target].upcase}"
+        "https://market.coss.io/api/trades/#{args[:base].upcase}_#{args[:target].upcase}"
```
wongy91, line 8: this should not gives the api link. It should gives the link where user can operate the trade.

In coss's case, there's no public viewing of trade page unless you are signed-up. Also signed-up is now invitation-only. So, we return `nil`

lib/cryptoexchange/exchanges/coss/services/pairs.rb
```diff
@@ -10,8 +10,8 @@ def fetch
         end
 
         def adapt(output)
-          output['result'].map do |value|
-            base, target = value['MarketName'].split('-')
+          output.map do |(market_name, value)|
```
wongy91, line 13: same here, no brackets required. `_value` for unused variable.

lib/cryptoexchange/exchanges/coss/market.rb
teekenl: ah i see

lib/cryptoexchange/exchanges/coss/services/market.rb
teekenl: 👍

lib/cryptoexchange/exchanges/coss/services/pairs.rb
teekenl: 👍

wongy91:
LGTM 👍

PR #2103, opened 2020-05-29
update stakecube api

wongy91:
update stakecube api endpoint


lib/cryptoexchange/exchanges/stake_cube/services/trades.rb
```diff
@@ -17,7 +17,7 @@ def adapt(output, market_pair)
             tr.base      = market_pair.base
             tr.target    = market_pair.target
             tr.type      = trade['type']
-            tr.price     = trade['price']
+            tr.price     = trade['PRICE']
```
teekenl, line 20: any reason they are just updating this attribute to be capital case?

wongy91: not sure why, but it's in their DOC as well.

PR #2104, opened 2020-05-29
Update BuyUCoin API

teekenl:
Update BuyUCoin API and its new parameters.
- [x] I have added Specs

```
```


lib/cryptoexchange/exchanges/buyucoin/services/market.rb
```diff
@@ -4,29 +4,17 @@ module Services
       class Market < Cryptoexchange::Services::Market
         class << self
           def supports_individual_ticker_query?
-            false
+            true
```
tmlee, line 7: Could use the `false` case with https://api.buyucoin.com/ticker/v1.0/liveData ?

lib/cryptoexchange/exchanges/buyucoin/services/market.rb
```diff
@@ -36,11 +24,11 @@ def adapt(output, market_pair)
...
-          ticker.high      = NumericHelper.to_d(output['high_24'])
-          ticker.low       = NumericHelper.to_d(output['low_24'])
-          ticker.change    = NumericHelper.to_d(output['change'])
-          ticker.last      = NumericHelper.to_d(output['last_trade'])
-          ticker.volume    = NumericHelper.to_d(output['vol'])
+          ticker.high      = NumericHelper.to_d(output['h24'])
+          ticker.low       = NumericHelper.to_d(output['l24'])
+          ticker.change    = NumericHelper.to_d(output['c24'])
+          ticker.last      = NumericHelper.to_d(output['LTRate'])
+          ticker.volume    = NumericHelper.to_d(output['LBVol'])
```
tmlee, line 31: cross checked with https://trade.buyucoin.com/markets?USDT-BTC it looks like the volume collect refers to `v24`

lib/cryptoexchange/exchanges/buyucoin/services/market.rb
teekenl: yes, we can definitely because initially, I saw they able to provide individual data so it might be better to just use it instead of querying them all.

lib/cryptoexchange/exchanges/buyucoin/services/market.rb
teekenl: Ah i see

lib/cryptoexchange/exchanges/buyucoin/services/market.rb
wongy91: batch query is good, because we do store them in cache. 50 ticker calls = 1 request to the api.

